PR #144, opened 2014-06-13
Coveralls.io

singingwolfboy:
Measure test coverage, report results to [Coveralls](https://coveralls.io/), display a badge in the README. Note that this depends on the tests measuring code coverage -- there's a pull request for that here: https://github.com/pyinvoke/invocations/pull/5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ install:
   - inv --list
   # Install remaining dev requirements (test runner, etc)
   - pip install -r dev-requirements.txt
+  # Install coveralls to report coverage info to Coveralls.io
+  - pip install coveralls
 script:
   # Main test suite
   - inv test
@@ -31,6 +33,8 @@ script:
   - pip install . # NO -e!
   - inv -l # Sanity check
   - flake8
+after_success:
+  - coveralls
 notifications:
   irc:
     channels: "irc.freenode.org#invoke"
--- a/README.rst
+++ b/README.rst
@@ -1,3 +1,5 @@
+|build-status| |coverage-status|
+
 Invoke is a Python (2.6+ and 3.2+) task execution tool & library, drawing
 inspiration from various sources to arrive at a powerful & clean feature set.
 
@@ -50,3 +52,8 @@ http://pyinvoke.org. Post-install usage information may be found in ``invoke
 You can install the `development version
 <https://github.com/pyinvoke/invoke/tarball/master#egg=invoke-dev>`_ via ``pip
 install invoke==dev --allow-unverified invoke``.
+
+.. |build-status| image:: https://travis-ci.org/pyinvoke/invoke.svg?branch=master
+   :target: https://travis-ci.org/pyinvoke/invoke
+.. |coverage-status| image:: https://coveralls.io/repos/pyinvoke/invoke/badge.png
+   :target: https://coveralls.io/r/pyinvoke/invoke
--- a/tasks-requirements.txt
+++ b/tasks-requirements.txt
@@ -1,5 +1,5 @@
 # Need invocations lib to load internal tasks module (for now)
--e git+https://github.com/pyinvoke/invocations#egg=invocations
+-e git+https://github.com/singingwolfboy/invocations@test-coverage#egg=invocations
 # Need self to, you know, run tests / load into tests / etc.
 -e .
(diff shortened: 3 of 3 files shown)


singingwolfboy:
@bitprophet ping

justinabrahms:
I'd (humbly) offer checking out codecov.io. It will report on how a particular PR is covered, in addition to what coveralls does, which is report on how the PR affects coverage.

I'm much less interested in (+0.02% coverage) than I am (This PR is 48% covered).

bitprophet:
a) I've been poking more at vanilla coverage.py lately and it is real nice, so this topic is on my mind
b) I agree with @justinabrahms that coveralls' output (as seen over on paramiko/paramiko for the last, like, yearish) is really not very helpful, I should check out codecov.io sometime.

bitprophet:
Double checked the current state of features for coveralls.io vs codecov and from their websites at least, codecov does still look more useful, coveralls continues to have the "+N% coverage" focus which is less useful.

Will test out actually using codecov and if I like how it looks, will add that to travis.

bitprophet:
Hrm, don't see any way to tell codecov to give flags to the report generation; currently my coverage stuff has to tell `coverage html` to `--include invoke/*` and `--omit invoke/vendor/*`, otherwise it thinks my suite only covers like 15% because it doesn't test most of the vendored material.

Similarly I don't see a way to give this info to coveralls.io.

Digging further to see why I opted to do the filtering at html-generation time, maybe I'll get lucky and there's a way to do it at coverage-generation time instead. Otherwise this will be kind of a no-op and I'll just have to jury-rig something stupid like "generate coverage report, then inspect it for the coverage percentage".

bitprophet:
Sweet, apparently at the time I was ignorant of the use of `.coveragerc` or it didn't exist - my suite needs to run under a nosetests wrapper, so `coverage run` isn't a good option. But `nosetests --with-coverage` honors `.coveragerc` so I think I am in business.

bitprophet:
Test upload seems to work reasonably well: https://codecov.io/github/pyinvoke/invoke?ref=07a956e789c25b721bd08442624b60c70efd47a8

It's like 0.5% less coverage than native `coverage html` reports, and this seems to be because it's only seeing about 1/2 of the branches - `coverage html` shows 584 branches but codecov only shows 289.

The 1/2-ness is constant across all covered files, which is...pretty bizarre. Can't find much online about this besides a probably-not-related twitter convo: https://twitter.com/glenathan/status/603883608651264000

That said, when it represents such a small amount-of-coverage for now, and only seems to affect branch coverage which is itself always a dicey proposition, Good Enough For Now.

justinabrahms:
Glad to hear that. The founder there is pretty responsive, in my
experience. Consider messaging him.

On Monday, December 21, 2015, Jeff Forcier <notifications@github.com> wrote:

> Test upload seems to work reasonably well:
> https://codecov.io/github/pyinvoke/invoke?ref=07a956e789c25b721bd08442624b60c70efd47a8
>
> It's like 0.5% less coverage than native coverage html reports, and this
> seems to be because it's only seeing about 1/2 of the branches - coverage
> html shows 584 branches but codecov only shows 289.
>
> The 1/2-ness is constant across all covered files, which is...pretty
> bizarre. Can't find much online about this besides a probably-not-related
> twitter convo: https://twitter.com/glenathan/status/603883608651264000
>
> That said, when it represents such a small amount-of-coverage for now, and
> only seems to affect branch coverage which is itself always a dicey
> proposition, Good Enough For Now.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pyinvoke/invoke/pull/144#issuecomment-166475245>.
>


-- 
Sent via mobile phone.

bitprophet:
Good to know, thanks again @justinabrahms!

Re: other bits in this PR, codecov's equivalent status badge can be found on https://codecov.io/github/pyinvoke/invoke/features/badges so I'll work that in, and then preserve the Travis status badge too cuz why not :)

bitprophet:
Welp Alabaster 0.7.7 is now out with Codecov button support :3

bitprophet:
Set CodeCov PR adjusted-line coverage to 90% for now, partly cuz that'd be nice, partly because I assume lots of PRs will fail and I want to see what failure looks like initially :3

PR #155, opened 2014-06-23
Flag values should shouldn't be parsed as flags

### LINKED ISSUE #154: String arguments within quotes shouldn't be parsed
I've run into a bug where values within a string argument are aggressively parsed by the CLI parser. Example:

``` python
# tasks.py
from invoke import task

@task
def foo(bar=False, baz=''):
    print 'bar:', bar
    print 'baz:', baz
```

Current output:

```
$ inv foo --baz="--bar"
bar: True
baz: 
```

Expected output:

```
bar: False,
baz: --bar
```

A workaround is to insert a space at the beginning of the string argument, but this is undesirable:

```
$ inv foo --baz=" --bar"
bar: False
baz:  --bar
```

pmdarrow:
- String arguments that look like flags should no longer be parsed. Fixes #154. 
- Upgrade `releases` to get builds going again.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,6 +1,6 @@
 # Docs
 Sphinx>=1.1.2
-releases>=0.6.1
+-e git+https://github.com/bitprophet/releases#egg=releases
 # Testing (explicit dependencies to get around a Travis/pip issue)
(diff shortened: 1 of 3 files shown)


joystein:
Maybe close this if it might be solved and its 8 years old?

pmdarrow:
Done 👍

Issue #132, opened 2014-04-03
List-type arg values


bitprophet:
Strongly related to #9, how to handle list values.

In #9, I assumed they would be contiguous values on the CLI with a separator, e.g. `inv mytask --listarg list;val;here`.

However, a common pattern for existing arg parsing libs is to give a flag >1 time and have the results tallied into a list, e.g. `inv mytask -l list -l val -l here`. E.g. this is how argparse/optparse do it, IIRC.

Pros of that latter approach:
- No need to deal with overridden meaning for shell meta-characters (many candidate delimiters have shell meaning.)
  - Counterpoint: commas work pretty good...
- No need to deal with escaping (always a PITA.)
- Minority use case: aping/wrapping pre-existing tools which work this way (e.g. using Invoke as a lightweight argument preprocessor before calling a Java process.)

Cons:
- Harder to implement, requires the parser to be more complicated :(
- Some parsers treat multiple values for the same flag as overrides, so in the above example we'd get `-l here` "winning" as a string, not list, value. Some users may expect or desire this behavior, e.g. when stitching together args from various "levels" of an invoking program.
- When you _are_ intending to give a list in a single invocation, typing out `mytask --listarg a;b;c` is much faster than `mytask -l a -l b -l c`.

Either way:
- Do we perform sub-list transformation, i.e. do we turn `mytask --listarg 1,2,3` into a list of ints automatically? Which again ties back into #9.

bitprophet:
(3 years later...) I've got a POC of this for the base case (list of strings) working, where one manually indicates which arguments should be list-type instead of string-type - similar to how positional and optional args are configured. E.g.:

```
@task(iterable=['listy'])
def mytask(c, listy=None):
    print(listy)
```

and then:

```
$ inv mytask --listy foo --listy bar
['foo', 'bar']
```

Need to backfill more tests & figure out some of the corner/edge cases:

- [ ] I've only got a high level CLI test atm so need a handful more for Parser, Argument at least
- [ ] Needs explicit mention in the docs (the autodoc for `@task` is not sufficient.)
- [ ] Make sure it meshes well with no-default-value task function arguments
- [ ] Worth trying to handle other types of list members? I can't see lists of ints or bools, for example, being a very common use case so it's probably okay to punt on that for now...
- [ ] This may exist already for the positional/optional features (in which case I may punt), but, the case where a user typos the argument name in `@task(iterable=[names...])` - right now it'll probably explode messily instead of usefully?
[...]

PR #181, opened 2014-09-28
Make an argument's value to a list if there are repeating value settings...

drakeguan:
This is for #132, List-type arg values.
1. Prevent ParseMachine's property `waiting_for_flag_value` from stopping repeating values.
2. In Argument.set_value(), if the `value` is not empty, make it as a list and append the given value.

It's a really quick hack, I know, and look forward your feedback and idea because I really like to make this feature work ASAP :)


bitprophet:
Ended up taking my own stab at this recently which handles things a bit differently (merging soon.) Thanks though!!

Issue #179, opened 2014-09-09
Option to turn off coloring in run() echo


ybendana:
It would be nice to have a kwarg for run() to turn off the echo coloring.

jhermann:
I likely will provide a `--color=no|auto|yes` patch at some time when I come around to it, with configurable themes, and at least `dark` and `light` (possibly based on `vim` themes).

SkiftCreative:
Working on this now.

halloleo:
Using `--color=...` with various options is of course better than a simple `--no-color`, but both would be a great improvement,

PR #189, opened 2014-10-18
Added nocolor flag to runner.run()

ybendana:
Fix for issue #179.


halloleo:
Just wanted to come back to this as well: "Is there any chance of this or something like it getting merged? The hardcoded color code here is a poor default–it's likely to match or nearly match the background color for light color schemes."

PR #190, opened 2014-10-28
Make invoke invokable with `python -m invoke`

techtonik:


neozenith:
This was eventually resolved with #444 in what I suspect later became v0.18.

Issue #170, opened 2014-08-20
Easy execution of tasks from within Python itself


bitprophet:
Until now the focus has been mostly on CLI execution, but calling programmatically has always been on the table - couldn't find a good ticket so this must be it :) (#112 is sort of a feature request for this but it's more of a support thread, I'll roll it into here instead.)

Currently there are no great options for calling a task from another task:
- The "main" way to execute presently is `Executor(<Collection obj>).execute('name')`, which requires one to have access to the right `Collection` instance.
  - Especially if that `Collection` is the one the task is itself a part of, this is basically impossible with the current API.
- Executing a task that lives in an imported collection is easier, one can simply do `that_collection['taskname'](args)`, however:
  - it's still a little fugly (if not a ton)
  - more importantly, it has annoying edge cases when using contextualized tasks - the context objects are different from each other so you can't just hand yours into the other task (though this feels like it should be allowed!)

Rambling thoughts about how to smooth over the above problems:
- Having some focus `Collection` when executing is still a must at some point - Fabric's `execute('name')` can't work here because it relies on implicit global state.
- The connection between a task's inner block and the Collection it lives in, must be via the `Context` handed in (or some other argument, but a single forced argument is plenty...).
  - Feels like more unification of (or smoothing over of access to) `Context` and `Collection` might help here.
  - At the very least we need a way to easily call a task with the "right" collection-as-configuration tree - similar to what the CLI mechanisms do now. Which may mean shoving more of that code into `Collection` or `Context`.
- The previous bullet is more fuel for removing non-contextualized tasks (#114)
  - However, the non-contextualized route just means you can't easily call contextualized tasks from non-contextualized ones (w/o making your own `Context` - which is doable). You can always just treat everything like a regular module-of-functions and call other nearby task functions straight up.

Brainstorm of use cases/scenarios:
- Call one top level task from another
- Call a subtask from a top level task
- ~~Call a top level task from a subtask~~ (this doesn't actually make any sense, by definition a subtask has no idea that it IS a subtask, at best it can be expected to know about its siblings)
- Call a subtask from another subtask in the same module (though ideally this is exactly the same as the first bullet)
- Call a top level task as a library call (eg in the shell)
- Call a subtask as a library call
- (Potentially a new ticket) Call one or more top level tasks, handing them the results of the parser - basically, control over the existing hardcoded-ish CLI module.

jhermann:
NOT having read the long thread, I'm still throwing #223 in the ring. :smile:

[...]

Issue #114, opened 2014-02-11
Consider making contextualization the default


bitprophet:
Occasionally catch myself doing the following:

``` python
from invoke import Collection, task

# stuff

@task
def foo(ctx, blah, otherblah):
    ctx.run("whatever")

ns = Collection(foo)
```

Which leads to a nice `'test' did not receive all required positional arguments!` error, because `task` != `ctask`. Also I think part of an FAQ mentions this (!)

If it's something I do myself cuz I haven't written a new Invoke task in a month...it's gonna bite users. The un-contextualized use case is and should be the minority; therefore I really should:

* [ ] Update docs to describe a "contextualized by default" setup, including removing mention of `@ctask`.
    * Maybe be hardcore and just don't even bother with uncontextualized tasks? Eh. Leaving the flag in allows people to rebind the name to an inverted-ctask if they really care about not using the context.
* [ ] Update tests to match
* [ ] Add new tests enforcing a useful "Did you forget a leading context argument?" add-on to the above error message, probably.
* [ ] Implement.
* [ ] Update all my stuff (whee! Good thing most of it is in Invocations...)
* [ ] Add big honkin' backwards incompatibility note to teh changelog (maybe a good new Releases feature?)

KuangEleven:
Working on removing non-contextualized tasks

thebjorn:
Perhaps a stupid question...?  but if all tasks need a first argument of a specific type, and it's a source of error that users forget this argument.. Wouldn't it be sensible to define a task as a subclass of `Task`, since then the first argument would be `self` which when missing is much more likely to be picked up by linters/IDEs?  I know you can define a class as a task already (e.g. https://github.com/datakortet/dk-tasklib/blob/master/dktasklib/lessc.py) and even though it works well it feels a little hacky..

bitprophet:
@thebjorn It's true that making them class definitions would make that line up a bit better. The problem is, this type of tool wants to be as DSL-like & lightweight as possible, so any additional boilerplate drives it farther from that goal.

The use of a context object is already a moderate boilerplate addition (one I judged worthwhile because of the large gains in explicitness, architectural sanity & debug friendliness). Adding class-def lines & extra indent levels strikes me as not quite worth the tradeoff.

[...]

PR #199, opened 2014-12-09
Make Task ok with MethodType tasks

sophacles:
This updates the argspec to allow for the body of the task to be a
method from an object instance, rather than just a function or a
**call**.  For the most part it is pretty straight forward, however the
way the executor._call_objs() test method set things up created a
wierdness in light of accepting methods.

Blindly poping the "first" argument of mock.**call** pretended self
was the ctx arg - not OK.  Rather we needed to create some simple 
task functions that proxied to a Mock object to allow for proper 
semantics.

### DIFF
--- a/tests/tasks.py
+++ b/tests/tasks.py
@@ -153,6 +153,25 @@ class Task_(Spec):
         def can_override_name(self):
             eq_(Task(_func, name='foo').name, 'foo')
 
+    class call_types:
+        def can_be_functions(self):
+            @task
+            def a_task():
+                return 10
+            eq_(a_task(), 10)
+        def can_be_classes(self):
+            class ATask(object):
+                def __call__(self):
+                    return 10
+            a_task = task(ATask())
+            eq_(a_task(), 10)
+        def can_be_methods(self):
+            class AnObj(object):
+                def foo(self):
+                    return 10
+            a_task = task(AnObj().foo)
+            eq_(a_task(), 10)
+
     class callability:
         def setup(self):
             @task


sophacles:
Ping @bitprophet  - added some unittests in 'tests/tasks.py' for this. The `can_be_method` fails in master, but passes in this branch. Not sure why the overall travis testing fails, but it appears unrelated to the tox testing, an on an issue different from the allowed callables for tasks. 

Also merged master into my branch for easy handling on your end :)

Where are your thoughts on this one?

bitprophet:
When I originally thought about this use case a tiny bit, I remember wondering whether the classes containing method tasks should themselves be Context subclasses, so that `self` == `ctx`.

This would not only make it easier to migrate to classes from contextualized functions in a module, but it means less boilerplate (two required args in every method == kinda gross) especially pending #114.

What are your thoughts on this re: your use case, @sophacles?

sophacles:
Regarding the aesthetics of `def taskname(self, ctx, ...)` I agree, it's a bit ugly. I'm not sure it's ugly enough to get rid of methods as tasks.

I understand the concerns about weird mutability issues between the object and the context - and I'm not sure how to avoid them easily. At least on the object side. I still think immutable contexts in the long run is a good way to go. On the object side it's sort of hard to do other than "don't do that" in the docs - something I think is problematic.

However also note - that having somewhere to keep long running state is trivial in python, so people are going to want and figure out how to do it - if nothing else module level variables. I even use that  sometimes with invoke now :/.

From the two different use cases I use based on a vendorized invoke with this patch:

(1) Upstart service management: When I made `class UpstartService` - and made my 20 different upstart services inherit from this class with a few special case overrides - I reduced about 400 lines of repeated boilerplate code. Further, the last couple of updates I did to the base class were about 3 hours faster (each!) to make happen, and the amount of debugging on all my deployment tests was far less. As a bonus - my `--list` output is far cleaner and more consistent, the users like that a lot.  I have similar results with my `uwsgi` deployment scripts.

These are completely simple to write in idempotent ways, and there is no temptation to keep state hiding out in the instantiated objects. A fresh instantiation each time would pose no problem.

(2) For the case of keeping around state between tasks - I have a couple of test classes that keep state and set variables for long runs in functional testing. The variables are things like PID, certain random generated values (e.g. customer ids, input strings, etc) that can be gotten other ways, but having them stored in variables is super convenient (vs writing to a tempfile, then retrieving the values e.g.). Those classes expose a few methods as tasks, and generally I functions/tasks instantiate the class and use that object for  a single run - but when working out what tests to do, I'll invoke the class directly with a few tasks, keeping state between them.

This use is convenient, and fresh instantiation between classes would cause issues with current use, however it is also something I'd rework if needed for keeping case 1 around.

invoke/tasks.py
```diff
@@ -131,15 +131,28 @@ def argspec(self, body):
...
+            spec = inspect.getargspec(body)
+            arg_names = spec.args[:]
+        elif isinstance(body, types.MethodType):
+            spec = inspect.getargspec(body)
+            arg_names = spec.args[1:]
+        else:
+            spec = inspect.getargspec(body.__call__)
+            arg_names = spec.args[1:]
+
+        #func = body if isinstance(body, types.FunctionType) else body.__call__
```
imhotepbot: * E265 block comment should start with '# '

invoke/tasks.py
```diff
@@ -131,15 +131,28 @@ def argspec(self, body):
...
+        #spec = inspect.getargspec(func)
```
imhotepbot: * E265 block comment should start with '# '

invoke/tasks.py
```diff
@@ -131,15 +131,28 @@ def argspec(self, body):
...
+        #arg_names = spec.args[:]
```
imhotepbot: * E265 block comment should start with '# '

tests/executor.py
```diff
@@ -74,18 +74,26 @@ def calls_default_to_empty_args_always(self):
 
         def _call_objs(self, contextualized):
             # Setup
-            pre_body, post_body = Mock(), Mock()
+            pre_mock, post_mock = Mock(), Mock()
+            def pre_body(a1, *args, **kw):
```
imhotepbot: * E301 expected 1 blank line, found 0

tests/tasks.py
```diff
@@ -153,6 +153,25 @@ def name_defaults_to_body_name(self):
...
+            def a_task():
+                return 10
+            eq_(a_task(), 10)
+        def can_be_classes(self):
+            class ATask(object):
+                def __call__(self):
+                    return 10
+            a_task = task(ATask())
+            eq_(a_task(), 10)
+        def can_be_methods(self):
```
imhotepbot: * E301 expected 1 blank line, found 0

tests/tasks.py
```diff
@@ -153,6 +153,25 @@ def name_defaults_to_body_name(self):
         def can_override_name(self):
             eq_(Task(_func, name='foo').name, 'foo')
 
+    class call_types:
+        def can_be_functions(self):
+            @task
+            def a_task():
+                return 10
+            eq_(a_task(), 10)
+        def can_be_classes(self):
```
imhotepbot: * E301 expected 1 blank line, found 0

bitprophet:
Going to declare this the spiritual successor of fabric/fabric#573 - I still have to carefully read all this great work by @sophacles in depth but the tl;dr of "let's find a way to bridge `Collection` and methods-as-tasks" is clearly being tackled here.

I had been kicking the idea around in my head before I remembered this ticket existed, and had sorta globbed onto an idea of `Collection.from_class` that extracts tasks from a class the same way `Collection.from_module` extracts tasks from a module. That may still be a useful addition to the work here.

Another random idea is to have a superclass or metaclass perform similar functionality - define your class-of-tasks, and then simply reference it (or an instance of it?) as if it were a valid Collection while building your collection/task trees.

----

Also want to make sure I link this to #170 since that's related - ideally "just call `self.foo()`" will prevail and #170 will be relegated to "regular" non-class-borne tasks that don't already have a baked-in `self` to refer to. But hard to say offhand w/o some thought/experimentation.

bitprophet:
Sidelining into #347 because I hate overloading PRs. I'll hopefully be poking this issue in the near future (after fabric 2 is alpha ready).

PR #157, opened 2014-06-28
Add flake8 testing to tox

Ivoz:
Might also consider adding support for python 3.4 in tox.ini & .travis.yml soon?

PR #212, opened 2015-03-04
rough flake8 integration

collinanderson:


dev-requirements.txt
```diff
@@ -7,5 +7,6 @@ alabaster>=0.6.0
 nose==1.3.0
 spec>=0.11.1
 mock==1.0.1
+flake8
```
bitprophet: Can you stick the latest stable release number in here as an `==`? thanks!

invoke/monkey.py
```diff
@@ -1,6 +1,9 @@
 # Fuckin' A.
 
-import select, errno, os, sys
+import errno
```
bitprophet: Unclear why this file is being touched, it's gone in HEAD. Can you try rebasing on latest HEAD perhaps?

invoke/parser/parser.py
```diff
@@ -88,7 +88,7 @@ def parse_argv(self, argv):
             # flag value and still in valid parsing territory (i.e. not in
             # "unknown" state which implies store-only)
             if not machine.waiting_for_flag_value and is_flag(token) \
-                and not machine.result.unparsed:
+                    and not machine.result.unparsed:
```
bitprophet: Is 8-space indents on continuations a flake8 thing? It's not my preferred style :( (Honestly, I thought flake8 would complain about backslash continuations entirely, parentheticals are better anyway).

setup.cfg
```diff
@@ -0,0 +1,4 @@
+[flake8]
+exclude=invoke/vendor,sites,tests
```
bitprophet: Any particular reason to exclude `tests`?

setup.cfg
```diff
@@ -0,0 +1,4 @@
...
+ignore=E124,E125,E128,E261,E302,E402,F401,W503,E712,F821
```
bitprophet: I should look these up to see what they are, am not a linting expert.

bitprophet:
Thanks for poking at this! See line comments :)

bitprophet:
Also linking to #157 which this kinda supersedes.

invoke/parser/parser.py
collinanderson: It just doesn't want the conditional indented at the same level as the code.

setup.cfg
collinanderson: Yeah, I figure start with a lot of them ignored. If individual ones get fixed they can be removed from this list.

invoke/parser/parser.py
collinanderson: E129 visually indented line with same indent as next logical line

dev-requirements.txt
collinanderson: done

invoke/monkey.py
collinanderson: Ahh. Yup.

invoke/parser/parser.py
collinanderson: This is now ignored under E125 for now.

setup.cfg
collinanderson: I force-pushed a change that now includes tests too. I just didn't want the diff to be too big :)

setup.cfg
collinanderson: Basically, you can remove one from the list, then run flake8 to see what lines they apply to.

bitprophet:
Thanks for making those updates. Annoyingly, I didn't notice last time that there's apparently a rule for double spaces before line comments (e.g. `foo = blah  # lol`) - this is another spot where I personally diverge from core PEP8.

Went and checked on all the stuff ignored/not ignored to see if I agree (I'd like to choose a single set of these rule ignores so I can apply them eventually to all of my projects :))...results:

* As above, I'm thinking we ignore `E261` (two spaces before line comments).
* Why ignore `F401` (module imported but unused)? When I run w/ it un-ignored I see 90% correct catches of unused imports (boy I should've put linting in before! thanks again for the PR...), with a handful of `__init__.py` convenience imports that we can presumably explicitly ignore.
* I don't see `W503` in pep8's docs, what's it do?
* Where are we violating `E711` and `E712`? Offhand I'd agree with those errors if they popped up. If I remove those from a copy of your branch, I see a handful of alerts and they all seem legit (where we use `==` instead of the arguably more correct `is`).
* `F811` and `F821` also sound like rules I'd agree with if they alerted, and as above, if I remove them it does find some things I would consider errors :)

collinanderson:
Don't worry, I added two space comments in the second change which is why you didn't catch it the first time.

I originally ignored all of the import cleanup just to keep the diff small enough. Here it is with all of the import cleanup.

bitprophet:
Looks good, thanks again! I might tweak that ignores list more in future but this is a great start.

PR #222, opened 2015-03-20
README.rst: Syntax highlight Python example

msabramo:
I like da colorz


bitprophet:
Don't see why not, thanks! I'm pretty lazy and am used to only viewing that snippet via the official website, http://www.pyinvoke.org/, which is Sphinx, which defaults to Python code highlighting for all blocks that appear to be valid for it. But it'd be nice to look better on Github/etc too.

PR #224, opened 2015-03-23
Add fish completion script

jaimeMF:


bitprophet:
Sweet, thanks!

jaimeMF:
Thanks!
Could you change my name in the changelog to `Jaime Marquínez Ferrándiz`?, `Marquínez` is my first surname (in Spain we use two surnames, [wikipedia article](https://en.wikipedia.org/wiki/Spanish_naming_customs))

bitprophet:
I'd been wondering why so many contributors with Spanish-sounding names always had their "middle" name listed! Sorry about that.

jaimeMF:
I would also get confused, thanks for the fix.

Issue #210, opened 2015-02-26
Documentation: the example for configuration raises an AttributeError


mlindemann:
Page http://docs.pyinvoke.org/en/latest/concepts/configuration.html, the following code example...

```python
from invoke import Collection, ctask as task

@task
def clean(ctx, target=None):
    ctx.run("rm -rf {0}".format(target or ctx.sphinx.target))

@task
def build(ctx, target=None):
    ctx.run("sphinx-build docs {0}".format(target or ctx.sphinx.target))

ns = Collection(clean, build)
ns.configure({'sphinx': {'target': "docs/_build"}})
```
... returns: `AttributeError: 'Context' object has no attribute 'sphinx'`

jenisys:
Which version of ``invoke`` are you using ?
I first used invoke-0.9.0 from pypi and ran into similar problems.
When I used the Github repository tip (which also says its version is 0.9.0), the problems were gone. 
NOTE: I am not quite sure if they are completely gone, but they are solved in most cases.

mlindemann:
I was using the pypi version.

Now, I have followed your advice and installed from Git.

Configuring by namespace works, i.e. the code below executes properly:

```python
from invoke import Collection, ctask as task

@task
def clean(ctx, target=None):
    ctx.run("echo CLEANING {0}".format(target or ctx.sphinx.target))

@task
def build(ctx, target=None):
    ctx.run("echo BUILDING {0}".format(target or ctx.sphinx.target))

ns = Collection(clean, build)
ns.configure({'sphinx': {'target': "docs/_build"}})
```

But configuring by invoke.yaml or invoke.py files still raises an error:

```python
# file: .../deploy_tools/tasks.py
from invoke import Collection, ctask as task

@task
def clean(ctx, target=None):
    ctx.run("echo CLEANING {0}".format(target or ctx.sphinx.target))

@task
def build(ctx, target=None):
    ctx.run("echo BUILDING {0}".format(target or ctx.sphinx.target))

ns = Collection(clean, build)
#ns.configure({'sphinx': {'target': "docs/_build"}})
```
```yaml
# file: .../deploy_tools/invoke.yaml
sphinx:
    target: built_docs
```

```bash
$ invoke clean
Traceback (most recent call last):
  File "/data/webapps/tdd/invoke-master/invoke/config.py", line 55, in __getattr__
    return self._get(key)
  File "/data/webapps/tdd/invoke-master/invoke/config.py", line 97, in _get
    value = self.config[key]
[...]

PR #240, opened 2015-05-02
Fix name of mock test method

gtback:
assert_has_call is not an assertion method on Mock() instances, so none
of these tests had a chance of failing.

I discovered this while attempting to add a test for #210. An alternative would be to switch to using `assert_any_call`.

### DIFF
--- a/tests/config.py
+++ b/tests/config.py
@@ -38,34 +38,34 @@ class Config_(IntegrationSpec):
             # This is a bit funky but more useful than just replicating the
             # same test farther down?
             Config(system_prefix='meh')
-            load_yaml.assert_has_call('meh.yaml')
+            load_yaml.assert_has_calls([call('meh.yaml')])
 
         @patch.object(Config, '_load_yaml')
         def default_system_prefix_is_etc(self, load_yaml):
             # TODO: make this work on Windows somehow without being a total
             # tautology? heh.
             Config()
-            load_yaml.assert_has_call('/etc/invoke.yaml')
+            load_yaml.assert_has_calls([call('/etc/invoke.yaml')])
 
         @patch.object(Config, '_load_yaml')
         def configure_user_location_prefix(self, load_yaml):
             Config(user_prefix='whatever')
-            load_yaml.assert_has_call('whatever.yaml')
+            load_yaml.assert_has_calls([call('whatever.yaml')])
 
         @patch.object(Config, '_load_yaml')
         def default_user_prefix_is_homedir(self, load_yaml):
             Config()
-            load_yaml.assert_has_call(expanduser('~/.invoke.yaml'))
+            load_yaml.assert_has_calls([call(expanduser('~/.invoke.yaml'))])
 
         @patch.object(Config, '_load_yaml')
         def configure_project_location(self, load_yaml):
             Config(project_home='someproject')
-            load_yaml.assert_has_call('someproject/invoke.yaml')
+            load_yaml.assert_has_calls([call('someproject/invoke.yaml')])
 
         @patch.object(Config, '_load_yaml')
         def configure_runtime_path(self, load_yaml):
             Config(runtime_path='some/path.yaml')
-            load_yaml.assert_has_call('some/path.yaml')
+            load_yaml.assert_has_calls([call('some/path.yaml')])
 
         def accepts_defaults_dict(self):
             c = Config(defaults={'super': 'low level'})


bitprophet:
Ugh, the problem with "any attribute is cool on a mock". Thanks for catching this.

I think the actual correct replacement is `assert_called_with`, no? (That lets you just slap in actual args, not a `call`, which was the original intent in these tests.)

gtback:
I'm not sure that `assert_called_with` works; at least, it didn't when I tried it (but I'm not an expert `Mock`er at all).  From what I can tell, `assert_called_with` only checks the most recent call, but `load_yaml` gets called several times (for the system config file, the project config file, the user config file, etc.) . `assert_any_call` is a bit less restrictive, but seems to pass all the Linux tests. 

```
Expected call: _load_yaml('/etc/invoke.yaml')
Actual call: _load_yaml('/Users/gback/.invoke.yaml')
```

In this case, by looking at `load_yaml.call_args_list`, I see `call('/etc/invoke.yaml') in the call list; it just isn't last.

The Windows tests are still failing and I'm not sure the best way to generalize them (particularly for the user and system tests).

gtback:
But yes, having the assert contain just args rather than `call()` instances would be better. :wink:

bitprophet:
Ah yea, you're right, the intent was "any call" not "first call", and `assert_any_call` should do that. Let's go with `assert_any_call` then. YAY BIKESHEDDING

gtback:
I finally got a Windows environment set up where I can test this locally. The two tests that seem to be failing are: 

* `config.init.default_system_prefix_is_etc`
* `config.init.configure_project_location`

You've already mentioned that the "system prefix" doesn't make much sense on Windows. Also [here](https://github.com/pyinvoke/invoke/blob/83c82244e48844aa846214c7b3449e2396b5d9ab/invoke/config.py#L301). Would it make sense for the "system-level" config file to just be `C:\invoke.(yaml|json|py)`?

The "project_location" test seems to just be a hard-coded path separator ("/") that doesn't translate on Windows. Easy fix.

gtback:
The Windows tests on AppVeyor seem to be passing (*crosses fingers*), but Travis seems to be getting errors building virtualenvs. I don't think that's related to this code. Perhaps manually restart the Travis build?

bitprophet:
Hm nope that's my fuckup, I started experimenting with code coverage and probably forgot to flip some bits somewhere. EDIT: found it, it's Travis-specific which is why I missed it. Pushed fix, will hit rerun on your build in a bit. Thanks!

bitprophet:
Er right, we can't actually rerun THAT build as it's SHA pinned. If you merge master I think it'll automatically kick off a new build.

bitprophet:
Also, +1 on default being just `C:\`, though - isn't there some semi-global "system stuff" paths in Windows, some junk using `%` characters and such? IIRC the problem is just that those things change depending on Windows version - so it may still be more widely applicable to just use the system volume root after all.

gtback:
Completely off the top of my head, so may be wrong:

There's `%SYSTEMROOT%`, which I think is typically `C:\Windows\` ... but I don't think you want users dropping config files in there. It's more like `/usr/bin` than `/etc` (gross generalization). 
In reality, we should probably use `%SYSTEMDRIVE%` rather than `C:\`, but I'll leave that choice to people who are actually planning to actively use invoke on Windows. :smile:

I'll merge master to kick off a new travis build.

gtback:
It looks like the `coverage` that is missing is actually a task in `invocations.testing`. I don't see it anywhere in the history of https://github.com/pyinvoke/invocations/commits/master/invocations/testing.py. Perhaps something you haven't pushed?

https://github.com/pyinvoke/invoke/blob/c9f11121a4fadfb17fbcd225fd3baeeb03db6e9e/tasks.py#L5
https://github.com/pyinvoke/invoke/blob/c9f11121a4fadfb17fbcd225fd3baeeb03db6e9e/tasks.py#L91-L94

gtback:
You should be able to "restart" a Travis Build if you're signed in. It's useful if the failure depends on something outside of the code (such as the wrong version of `invocations` getting pulled).

bitprophet:
Ugh, yes, that task is local, sorry. The danger of having a bunch of interconnected `pip install -e` versions of things. Just pushed, will try to make travis rebuild after (which in this case, yea, would work; earlier, it was a change to dev-requirements in this repo, which did require a merge).

bitprophet:
Gross, travis (distinct from my testing on workstations) is really mad about something in the newer pty related code (so, not your fault). I'll have to figure that out.

PR #243, opened 2015-05-05
Pass in project directory when tasks module contains explicit namespace.

### LINKED ISSUE #234: Trouble loading the project-level configuration file
I'm using Invoke 0.10.1 in an virtual environment. I have an `invoke.yaml` file next to my `tasks.py` file, but it is ignored. Values from `~/.invoke.yaml` are loaded though. If i use `-f` it works too.

The `tasks.py` imports some tasks from other modules, but the root namespace is created and configured in the `tasks.py` file. I tried using a package for the tasks instead but to no avail.

Any hints?

And btw: if I use a package, do the `invoke.yaml` file needs to be in the package (next to `__init__.py`) or in the parent folder (tried both, none worked)?

gtback:
When a `tasks.py` module contains an explict Collection (named either `ns`
or `namespace`), the parent directory (used to find the project-specific
configuration file (e.g. `invoke.yaml`) is not passed in to the
Collection created from the factory method `Collection.from_module`.
This caused the file to not be found except when explicitly specified
with the -f flag. Compare to the similar Collection() invocation about
12 lines below this change.

Fix #234.

### DIFF
--- a/invoke/collection.py
+++ b/invoke/collection.py
@@ -163,7 +163,8 @@ class Collection(object):
                 # TODO: make this into Collection.clone() or similar
                 # Explicitly given name wins over root ns name which wins over
                 # actual module name.
-                ret = Collection(name or obj.name or module_name)
+                ret = Collection(name or obj.name or module_name,
+                                 loaded_from=loaded_from)
                 ret.tasks = copy.deepcopy(obj.tasks)
                 ret.collections = copy.deepcopy(obj.collections)
                 ret.default = copy.deepcopy(obj.default)


gtback:
I'm happy to add a test for this, but I'm not sure the best place to, since it relies on having both tasks.py and invoke.yaml in the same directory. Trying to solve this is what led me down the path to #240, but I don't think there's a good unit test for this problem. An integration test, perhaps?

bitprophet:
Naw, there was actually a regular semi-unit test (the ones in the CLI module are in a gray area between unit and integration) that can be copied to work perfectly for this, just needed a new fixture & a tweak to select it.

Issue #68, opened 2013-06-07
Consider option for disabling .pyc generation


brejoc:
Might be nicer of invoke to not generate a tasks.pyc.

brutus:
This sounds pretty neat. What's the state of this and <https://github.com/pyinvoke/invoke/commit/ea9321821bd59976179ef212450357e31892667a>?

bitprophet:
Is that commit part of an actual PR somewhere? If it is, github's not telling me.

brutus:
I just tired it again here: https://github.com/brutus/invoke/commit/e4240746404efbc1083bb12aa3f3f9c64c7ecb06

It still works for me. Just had a quick look though. The test run fine expect one that fails on the additional `--write-pyc` option in the help output as expected.

bitprophet:
Implemented, finally. Thanks again @brutus!

PR #244, opened 2015-05-07
Feature #68: Only build bytecode files if option flag is used

brutus:
See: <https://github.com/pyinvoke/invoke/issues/68>. Since I generally like clean / less noisy project directories I think it's okay to default to not writing bytecode files (`*.pyc`) unless the option flag is set. Invoke tends to be pretty fast anyway and the possible speedup trough the bytecode files might be neglect able.

Not sure about tie ins to the [configuration system](http://invoke.readthedocs.org/en/latest/concepts/configuration.html)…

### DIFF
--- a/invoke/cli.py
+++ b/invoke/cli.py
@@ -205,6 +205,12 @@ def parse(argv, collection=None, version=None):
             help="Enable debug output.",
         ),
         Argument(
+            names=('write-pyc',),
+            kind=bool,
+            default=False,
+            help="Enable creating .pyc files.",
+        ),
+        Argument(
             names=('echo', 'e'),
             kind=bool,
             default=False,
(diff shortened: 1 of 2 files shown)


bitprophet:
Thanks, added to milestone.

Re: configuration, I don't see any spots where we currently honor config files for CLI-oriented flags - most things impact `run()` (e.g. `--echo`) and thus aren't even inspected until the CLI module is done doing things.

We should probably honor this type of option (e.g. `root`, `no-dedupe`, `collection`, `debug`, etc - in addition to this new one) but I don't think it's a blocker here.

bitprophet:
Welp I let this sit so long it got obviated by updates to cli.py (now program.py). Manually porting this code & adding extra tests too n such. Will credit in changelog!

Issue #254, opened 2015-06-12
SyntaxError message appear in the case of python3 environment pip.


yoshiya0503:
[issue]
I think 'Invoke' is best choice for running tasks, but there are some week points about vendor.
For example, we install 'invoke' by 'pip3' in the case of using python3 virtual environment created by 'pyvenv' or 'virtualenv', SyntaxError message will appear.
In spite of the above message, Install has success.(however it looks misshapen...)

[cause]
AFAIK, the cause of the above issue is that both 'yaml2' and 'yaml3' in 'vendor' are installed.
Hopefully, managing 3rd party package should be replace more better method.
(using 'install_requires' of setuptools)
However, it is not so easy to replace and test. 
Then, first of all, either yaml2 or yaml3 should be required depending on the situation(python3 or python2).
This change will be very small and easy.

[pip install log]
Downloading/unpacking invoke
  Downloading invoke-0.10.1-py2.py3-none-any.whl (166kB): 166kB downloaded
Installing collected packages: invoke
*** Error compiling '/private/tmp/pip_build_root/invoke/invoke/vendor/yaml2/constructor.py'...
  File "/private/tmp/pip_build_root/invoke/invoke/vendor/yaml2/constructor.py", line 130
    except TypeError, exc:
                    ^
SyntaxError: invalid syntax

*** Error compiling '/private/tmp/pip_build_root/invoke/invoke/vendor/yaml2/reader.py'...
  File "/private/tmp/pip_build_root/invoke/invoke/vendor/yaml2/reader.py", line 158
    except UnicodeDecodeError, exc:
                             ^
SyntaxError: invalid syntax

*** Error compiling '/private/tmp/pip_build_root/invoke/invoke/vendor/yaml2/resolver.py'...
  File "/private/tmp/pip_build_root/invoke/invoke/vendor/yaml2/resolver.py", line 171
    re.compile(ur'''^(?:yes|Yes|YES|no|No|NO
                    |true|True|TRUE|false|False|FALSE
                    |on|On|ON|off|Off|OFF)$''', re.X),

                                             ^
SyntaxError: invalid syntax

*** Error compiling '/private/tmp/pip_build_root/invoke/invoke/vendor/yaml2/scanner.py'...
  File "/private/tmp/pip_build_root/invoke/invoke/vendor/yaml2/scanner.py", line 1427
    except UnicodeDecodeError, exc:
                             ^
SyntaxError: invalid syntax

Successfully installed invoke
Cleaning up...

[...]

PR #255, opened 2015-06-12
add exclude option in setup.py #254

yoshiya0503:
Hi team.

I created the issue and PR.
see #254

Best regards.

-------
Yoshiya Ito

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,12 +3,22 @@
 # Support setuptools only, distutils has a divergent and more annoying API and
 # few folks will lack setuptools.
 from setuptools import setup, find_packages
+import sys
 
 # Version info -- read without importing
 _locals = {}
 with open('invoke/_version.py') as fp:
     exec(fp.read(), None, _locals)
 version = _locals['__version__']
+
+# select yaml package depending on pip
+PY2 = sys.version_info[0] == 2
+PY3 = sys.version_info[0] == 3
+exclude = []
+if PY3:
+    exclude.append('*.yaml2')
+else:
+    exclude.append('*.yaml3')
 
 # Frankenstein long_description: version-specific changelog note + README
 long_description = """
@@ -29,7 +39,7 @@ setup(
     author_email='jeff@bitprophet.org',
     url='http://docs.pyinvoke.org',
 
-    packages=find_packages(),
+    packages=find_packages(exclude=exclude),
     entry_points={
         'console_scripts': [
             'invoke = invoke.cli:main',


bitprophet:
Oh neat, I didn't know about this feature of `find_packages`, this will definitely help a lot of folks who are being bitten by this issue. Thank you! Adding to release milestone.

yoshiya0503:
Thank you for your reply!
Any time! :)

bitprophet:
Side note, a user claimed in #258 that the pyyaml SyntaxError on install, wasn't present in master. I'm guessing that may be something else at work since we haven't merged this PR yet and I don't recall other recent changes that would affect the vendored pyyaml. But, worth verifying.

yoshiya0503:
Hi.

I verified the issue #258 following environment, and I reproduce #258 install error.
Mac OSX v10.10.4 (OS is so irrelevant about #258)
pip v1.5.6
python v2.7.9

I hope this information will help you.
Best Regards.

[root cause]
The root cause of #258 is that yaml3 is installed despite of using python2.7x virtualenv.
In fact, the following syntax is valid only python3.

class YAMLObject(metaclass=YAMLObjectMetaclass):
...

By merging this PR, it will be avoidable the install error.
However, as you said, there is also other cause.
That is to say, the cause of #258 is that users are using pip v1.5.6.
pip1.5.x compile the 3rd-party libs during install.(.pyc)
(the message of compiling appears to be using the pip1.5.x)
Then, yaml3(for python3) compiling caused the error in the case of using python 2.x of virtualenv. 

[easy solution]
try following.
(your virtualenv)> pip install --upgrade pip
and try again
(your virtualenv)> pip install invoke

[investigation log]
> virtualenv Invoke
New python executable in Invoke/bin/python
Installing setuptools, pip...done.

>  cd Invoke
> source ./bin/activate
> pip install -U pip==1.5.6
> pip --version
pip 1.5.6 from /Users/itouyoshiya/Documents/Python/Invoke/lib/python2.7/site-packages (python 2.7)

> pip install invoke
Downloading/unpacking invoke
  Downloading invoke-0.10.1-py2.py3-none-any.whl (166kB): 166kB downloaded
Installing collected packages: invoke
Compiling /Users/itouyoshiya/Documents/Python/Invoke/build/invoke/invoke/vendor/yaml3/__init__.py ...
  File "/Users/itouyoshiya/Documents/Python/Invoke/build/invoke/invoke/vendor/yaml3/__init__.py", line 284
    class YAMLObject(metaclass=YAMLObjectMetaclass):
                              ^
SyntaxError: invalid syntax

Successfully installed invoke
Cleaning up...

ivoflipse:
I'm currently having a similar problem where conda (the package manager) also compiles files from the yaml3 folder and thus fails my CI build. If I could make sure that it didn't get included in the Python 2.7 package to begin with, this could be prevented.

So it would be great if this got merged in, so I can easily keep creating conda packages for invoke

bitprophet:
Yup this is in the 0.12 milestone and I plan to cut 0.12 any day now (already kicked most of the stuff I can't easily knock out, to 0.13, but not this one!). Sorry it's taken a while!

ivoflipse:
Thanks for the heads up! Looking forward to 0.12 then

bitprophet:
For the record, I can't trivially recreate this problem with pip 7.1.2 from a Python 2 host environment (re: virtualenv creation). E.g. `virtualenv --python=python3 blah && source blah/bin/activate && pip install <master invoke path>` doesn't kaboom.

That's not to say I don't believe this problem exists, it's definitely gotten enough reports - just noting that I'm about to merge an updated copy of this PR but without being able to verify it. Gonna be relying on @yoshiya0503 @matheo @ivoflipse etc to verify it for me :) please check out latest master and let me know if it corrects the issue for you!

I'll probably release things as-is without waiting, but happy to cut another emergency bugfix release should I have incorrectly applied the fix.

ivoflipse:
I didn't get the problem using pip, but while building an invoke package using [conda](http://conda.pydata.org/docs/). It tries to compile all the .py files in the package, which obviously fails if Python 3 code is run while building for Python 2.

Having this filtering step would make my life easier :-)

PR #265, opened 2015-07-27
Build improvements

thedrow:
Added PyPy3 to the build, moved to container based infrastructure and enabled pip caching.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,8 @@
 language: python
+sudo: false
+cache:
+  directories:
+    - $HOME/.cache/pip
 python:
   - "2.6"
   - "2.7"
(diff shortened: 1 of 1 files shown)


thedrow:
Can you explain why the build fails so I can fix it?
Is it because PyPy3 is an implementation of Python 3.2?

bitprophet:
Thanks for this! The issue must be that our (travis.yml) test to skip doc-builds on 3.2 (Sphinx doesn't support it) isn't correctly triggering on PyPy3, I bet `TRAVIS_PYTHON_VERSION` isn't being set to `"3.2"` in that case (guessing "PYPY" is in there somewhere, or...something?)

Compare https://travis-ci.org/pyinvoke/invoke/jobs/72795107#L5996 to https://travis-ci.org/pyinvoke/invoke/jobs/72795111#L1027 - "real" Python 3.2 correctly skips the docs build step, but PyPy3 does not.

I don't see the build output stating what that value is, you'll want to make some test commits adding things like `- echo TRAVIS_PYTHON_VERSION: '$TRAVIS_PYTHON_VERSION'` or whatnot, so we can see what the value is.

bitprophet:
FTR we nixed 3.2 recently so that should be moot now. I merged and we'll see if Travis likes the sum total of the changes soon...

bitprophet:
OK, as semi-expected, PyPy3 fails because of Sphinx, but passes everything else: https://travis-ci.org/pyinvoke/invoke/jobs/102019458

Issue #259, opened 2015-07-10
Ability to specify environment variables to be used with the run function


fgimian:
Hey there, love invoke.

In fabric, we have `with shell_env(...)` so that a particular set of environment variables will be used in a particular scope, which apply to commands being run.

Is there any way to currently accomplish this with invoke?

I would really love to switch back to it, but I've reluctantly switched to paver until I can figure this out.

Thanks
Fotis

bitprophet:
There's no explicit support for this yet, but since Invoke is primary concerned with local-to-system shell commands, you should be able to modify `os.environ` yourself & the subprocess ought to inherit the current state of that data structure when you call `run`.

So this sort of thing ought to Just Work (tm):

``` python
import os
from invoke import run

my_vars = {'FOO': 'bar'}
os.environ.update(my_vars)
run("echo $FOO") # => should result in stdout of "bar"
```

That's not to say we won't add a 'helper' for this (i.e. a contextmanager that modifies `os.environ` only in a block) or take stronger ownership of the env presented to our subprocesses (e.g. making them default to an empty env, then offering something like `run("command", env={'a': 'dict'})`) so I'll leave this open as a placeholder for those concerns. Thanks!

fgimian:
Thanks so much for your reply @bitprophet, this will do for now.  I'm all in favour of allowing us to pass in the env kwarg.  The only con with the approach above is that it modifies os.environ for the entire script, one would need to save the original and restore it afterwards if they wanted to be a little neater :smile: 

A simple context manager would probably also do the trick, something like this (not tested yet):

```python
def environ(env):
    original_environ = os.environ.copy()
    os.environ.update(env)
    yield
    os.environ = original_environ
```

Anyways, thanks a lot for all your help and have a lovely weekend!

Cheers
Fotis

fgimian:
Yup, just tested this little solution and it does nicely :smile: 

```python
from contextlib import contextmanager
import os

@contextmanager
def environ(env):
    original_environ = os.environ.copy()
    os.environ.update(env)
    yield
    os.environ = original_environ

print os.environ.get('name')
with environ({'name': 'Invoke Me :o'}):
    print os.environ.get('name')
print os.environ.get('name')
```

Output:

```
➔ python testing.py 
None
Invoke Me :o
None
```

Cheers
Fotis

bitprophet:
[...]

Issue #67, opened 2013-06-07
Allow arbitrary shells


brejoc:
Make sure users can determine which local shell is used when executing commands.

----

Was: *Only Bash as shell?*

> I've only had a quick peek at the code, but I think the bash shell is hardcoded in the runner. Would be nice to let the user define the the shell in tasks.py, since some shells don't share the same syntax.

bitprophet:
Ran into an internal need for this in #152 (see final comment), so, poking now.

The tl;dr of current state of this issue:

* Configuration has been in for a long time now (this ticket is old) so not a problem
* PTY-driven `run` sets its own shell, which is `/bin/bash`, so that's easily updated
* non-PTY-driven `run` uses `subprocess`' own hardcoded default, `/bin/sh`, but that should be overrideable via the `executable` kwarg

bitprophet:
All done. Simply configure `run.shell` via any config method, or say `run('command', shell='/bin/fish')` or whatnot at runtime.

presidento:
Great, thank you @bitprophet!
(Didn't you want to add it to the 0.12.3 milestone?)

frol:
Just for your info, the implementation of this feature breaks Invoke on Windows: #345.

PR #280, opened 2015-09-29
specify environment variables in run()

philtay:
Ref: #259

Tests and docs after your approval and first round of comments.

### DIFF
--- a/invoke/runners.py
+++ b/invoke/runners.py
@@ -474,7 +474,9 @@ class Local(Runner):
                 # bash for now because that's what we have been testing
                 # against.
                 # TODO: also see if subprocess is using equivalent of execvp...
-                os.execve('/bin/bash', ['/bin/bash', '-c', command], env=self.env)
+                os.execve(
+                    '/bin/bash', ['/bin/bash', '-c', command], env=self.env
+                )
         else:
             self.process = Popen(
                 command,
--- a/invoke/runners.py
+++ b/invoke/runners.py
@@ -205,6 +205,7 @@ class Runner(object):
             runtime = kwargs.pop(key, None)
             opts[key] = value if runtime is None else runtime
         # TODO: handle invalid kwarg keys (anything left in kwargs)
+        env = kwargs.pop('env', None)
         # Normalize 'hide' from one of the various valid input values
         opts['hide'] = normalize_hide(opts['hide'])
         # Derive stream objects
@@ -219,6 +220,13 @@ class Runner(object):
             print("\033[1;37m{0}\033[0m".format(command))
         # Determine pty or no
         self.using_pty = self.should_use_pty(opts['pty'], opts['fallback'])
+        # Environment variables
+        self.env = os.environ.copy()
+        if env is not None:
+            if env:
+                self.env.update(env)
+            else:
+                self.env = {}
         # Initiate command & kick off IO threads
         self.start(command)
         encoding = opts['encoding']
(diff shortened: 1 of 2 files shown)


invoke/runners.py
```diff
@@ -466,13 +474,14 @@ def start(self, command):
                 # bash for now because that's what we have been testing
                 # against.
                 # TODO: also see if subprocess is using equivalent of execvp...
-                os.execv('/bin/bash', ['/bin/bash', '-c', command])
+                os.execve('/bin/bash', ['/bin/bash', '-c', command], env=self.env)
```
betterdiff: * E501 line too long (82 > 79 characters)

philtay:
Don't know exactly why the tests fail. On my PC tox doesn't report any problem. Anyway, please find below a simple task that shows the feature.

```python
from invoke import run, task

@task
def foo():
    run("env")
    print "************************************************************"
    run("env", env={'FOO':'BAR'})
    print "************************************************************"
    run("env", env={})
```

invoke/runners.py
```diff
@@ -219,6 +220,13 @@ def run(self, command, **kwargs):
             print("\033[1;37m{0}\033[0m".format(command))
         # Determine pty or no
         self.using_pty = self.should_use_pty(opts['pty'], opts['fallback'])
+        # Environment variables
+        self.env = os.environ.copy()
+        if env is not None:
+            if env:
+                self.env.update(env)
+            else:
+                self.env = {}
```
sophacles: This is a pretty big breaking change to current (and arguably expected) behaviour. If the user doesn't pass an environment, the child process is executed with no environment (empty dict). The general expected behaviour is that a child is spawned with the environment of it's parent unless otherwise specified.

This would probably be best as:

```
 # Environment variables
self.env = os.environ.copy()
if env:
    self.env.update(env)
```

philtay:
@sophacles thanks for your feedback. if the user doesn't pass an environment the child process is executed WITH the parent's environment. there is no breaking change. only if the user passes an empty dict the child process is executed with no environment.

my implementation is:

* non empty dict -> add these vars to the current env and execute the process
* empty dict (or something else that evaluates to False in boolean context) -> execute the process with no env
* if the user doesn't pass an env (or None) -> execute the process with the parent's env (i.e. current behavior)

sophacles:
@philtay Oh got it, I was a bit confused on that chunk of code. I understand the logic now - thanks for the clarification

Thinking further about it - I think perhaps the best thing to do would be mirror the `subprocess` environment logic, that is:

```python
                           # from subprocess.py line 1281 in python 2.7.10
                            if env is None:
                                os.execvp(executable, args)
                            else:
                                os.execvpe(executable, args, env)
```

Using values from `os.environ` should be left to the user to set up: there are too many edge cases otherwise:

* user wants to remove some variables from `os.environ`
* user wants to only have a specific environment (note #67 will also need to be completed before this case is fully available - setting shell to something like `/bin/env -i /bin/bash`) or alternately a runner with `shell=False` available.
* etc

sophacles:
Tangentially related to my comment above: the `invoke` code has:

```
                # TODO: also see if subprocess is using equivalent of execvp...
```

@bitprophet - the answer to this todo is: yes, subprocess uses `execvp` and `execvpe` when not on windows.

philtay:
@sophacles i agree with you. i'll wait for @bitprophet input before making changes.

bitprophet:
Went ahead with my own version of this today; credited in changelog though. Thanks!

PR #281, opened 2015-10-20
Update troves for moar python support

mattrobenolt:
Travis tests against 3.4 and 3.5, but the classifiers don't say they're compatible!

Also changed `wheel` -> `bdist_wheel` in the setup.cfg since `bdist_wheel` is the recommended way.


bitprophet:
thanks!

PR #278, opened 2015-09-28
A bunch of work for refactoring Program re #276

sophacles:
Took on a first pass at the refactoring I suggested in #276 to see how
much work it was. This commit does it. The major things:

* Create an `InvokeProgram` subclass of `Program`, for the `invoke` command.
* Rearrange how the arguments are handled a bit:
  + Move task related arguments to `core_args` in `InvokeProgram`
  + split argument handling out of `_parse`
  + Add a few functions for stages of argument handling
* remove references to argument values from `Program` for all arguments
  that are not in `core_args`
* fix up tests to run
* rearrange tests to differentiate between `Program` tests and
  `InvokeProgram` tests
  + new test file `invoke_program.py`
  + new `_utils` function expect_inv to run integration tests with the
  `InvokeProgram` subclass rather than the `Program` class

An important note here, even if this work is not the best direction to
take invoke is that there are a few places in program where core args
and task args are mixed together and expected. These cause problems for
developers creating their own "binaries" from `Program`:

* In `Program` the method `load_collection` expects args to have
  `args.root` and `args.collection` defined (whether or not they have a
  value)
* In `Program` the `config` property expects `args['no-dedupe']` to be
  defined

PR #285, opened 2015-11-14
Module needs to be converted to a Collection

matheo:
I was copy pasting the example and it didn't work
so reading more about namespaced I saw the Collection.from_module method to make it work.

Thanks a lot for this great stuff!!!!

### DIFF
--- a/sites/docs/concepts/library.rst
+++ b/sites/docs/concepts/library.rst
@@ -115,10 +115,10 @@ sake.
 Now the useful part: telling our custom `.Program` that this namespace of tasks
 should be used as the subcommands for ``tester``, via the ``namespace`` kwarg::
 
-    from invoke import Program
+    from invoke import Program, Collection
     from tester import tasks
 
-    program = Program(namespace=tasks, version='0.1.0')
+    program = Program(namespace=Collection.from_module(tasks), version='0.1.0')
 
 The result?
 


matheo:
Oh, I see this "Program" stuff is work in progress right?
Because i'm getting this error trying to execute subcommands :(

```
  File "/usr/local/lib/python3.4/dist-packages/invoke/vendor/lexicon/attribute_dict.py", line 4, in __getattr__
    return self[key]
  File "/usr/local/lib/python3.4/dist-packages/invoke/vendor/lexicon/alias_dict.py", line 80, in __getitem__
    return self._handle(key, None, single, multi, unaliased)
  File "/usr/local/lib/python3.4/dist-packages/invoke/vendor/lexicon/alias_dict.py", line 62, in _handle
    return unaliased(self, key, value)
  File "/usr/local/lib/python3.4/dist-packages/invoke/vendor/lexicon/alias_dict.py", line 74, in unaliased
    def unaliased(d, key, value): return super(AliasDict, d).__getitem__(key)
KeyError: 'list'

During handling of the above exception, another exception occurred:

  File "/usr/local/bin/theo", line 9, in <module>
    load_entry_point('Theo==0.1.0', 'console_scripts', 'theo')()
  File "/usr/local/lib/python3.4/dist-packages/invoke/program.py", line 246, in run
    self._parse(argv)
  File "/usr/local/lib/python3.4/dist-packages/invoke/program.py", line 324, in _parse
    if self.args.list.value:
  File "/usr/local/lib/python3.4/dist-packages/invoke/vendor/lexicon/__init__.py", line 23, in __getattr__
    return super(Lexicon, self).__getattr__(key)
  File "/usr/local/lib/python3.4/dist-packages/invoke/vendor/lexicon/attribute_dict.py", line 7, in __getattr__
    raise AttributeError(key)
AttributeError: list
```
Where "Theo" is my Program

sophacles:
Not sure if this specifically is handled, but in #278 there is a bunch of stuff to help the "to `Program`" conversion.

matheo:
I tried it but didn't run, any advice to make it work while @bitprophet gets his vacations to love invoke again? :)

sophacles:
So the biggest disconnect I ran into, and what seems to be one of the root causes for what you're encountering, is the current thinking behind the `Program` class. In master, the program class has a bunch of code specifically for the case where the binary used to execute tasks is called `invoke`. (note the PR is linked above separates this out into a class called `InvokeProgram`). This code handles things like the `--list` argument, and a few others with the underlying idea that a program may not want those `invoke` specific features. As you mention, it's a work in progress, so how it will all shake out in the end is up in the air, so I suggest that if possible you hold off on designing things around the current implementation of `Program`. 

If you choose to go forward with the knowledge that things are going to change, you may want to try my branch and using `InvokeProgram` instead of `Program`, but no guarantees that it will work for your case. It will be interesting to see what you find in exploration though, if you do take this route!

Also note: In master there are a few other bugs where the `invoke` specific logic above isn't fully factored away - this means anything other than `invoke` as the binary will break right now :(

Issue #276, opened 2015-09-25
Things with the new Program class


sophacles:
So there are a few things I've encountered with the new `Program` class. A few of them are bugs. A couple of them are usage things.

**Bugs:**

For the following bugs, please refer to this repository:  https://github.com/sophacles/inv-tests in the `inv_0.12 ` directory.

Quick overview: I create a module with a few library files, and main.py that defines a couple different program objects to serve as entry points into my task library. This module is installed into a virtual environment, along with the latest `invoke` from master. 

This is a reduction of my current invoke use case - a library of various task items and "ops-like" tasks, that are wrapped up into a few command line tools.

Here's the output I'm getting from this:

```
eheine@wirelessprvnat-172-17-31-229 ~/src/projects/invoke $ mkvirtualenv invtestNew python executable in invtest/bin/python2.7
# ...
Successfully installed invoke
Cleaning up...
(invtest)eheine@wirelessprvnat-172-17-31-229 ~/src/projects/invoke $ cd ../invthings/inv-tests/inv_0.12/lilprog 
(invtest)eheine@wirelessprvnat-172-17-31-229 ~/src/projects/inv-tests/inv_0.12/lilprog $ pip install -e .
# ---
Successfully installed lilprogtest
Cleaning up...
(invtest)eheine@wirelessprvnat-172-17-31-229 ~/src/projects/invthings/inv-tests/inv_0.12/lilprog $ myprog --list
No idea what '--list' is!
(invtest)eheine@wirelessprvnat-172-17-31-229 ~/src/projects/invthings/inv-tests/inv_0.12/lilprog $ otherprog --list
No idea what '--list' is!
(invtest)eheine@wirelessprvnat-172-17-31-229 ~/src/projects/invthings/inv-tests/inv_0.12/lilprog $ myprog subtask1
Traceback (most recent call last):
  File "/Users/eheine/envs/invtest/bin/myprog", line 9, in <module>
    load_entry_point('lilprogtest==0.1b1', 'console_scripts', 'myprog')()
  File "/Users/eheine/src/projects/invoke/invoke/program.py", line 246, in run
    self._parse(argv)
  File "/Users/eheine/src/projects/invoke/invoke/program.py", line 324, in _parse
    if self.args.list.value:
  File "/Users/eheine/src/projects/invoke/invoke/vendor/lexicon/__init__.py", line 23, in __getattr__
    return super(Lexicon, self).__getattr__(key)
[...]

PR #288, opened 2015-12-13
Added list to core_args and added check for no-dedup. Program was ass…

jordoncm:
…uming these would be a part of the flag structure which breaks creation of a Program.run where a namespace is passed in.

Consider the self contained python script ./foo

``` python
#!/usr/bin/env python

import sys

import invoke

@invoke.task
def build():
  print 'built!'

if __name__ == '__main__':
  namespace = invoke.Collection()
  namespace.add_task(build)
  program = invoke.Program(
    name='Foo',
    binary='foo',
    version='0.1',
    namespace=namespace
  )
  program.run(sys.argv)
```

`invoke.Program` would except because it assumes `self.args.list` was defined. The list operation makes sense to be a part of core_args in this use case. After fixing this `invoke.Program` excepts again assuming `self.args['no-dedup']` is defined.

### DIFF
--- a/invoke/program.py
+++ b/invoke/program.py
@@ -64,6 +64,12 @@ class Program(object):
                 help="Set default value of run()'s 'hide' kwarg.",
             ),
             Argument(
+                names=('list', 'l'),
+                kind=bool,
+                default=False,
+                help="List available tasks."
+            ),
+            Argument(
                 names=('pty', 'p'),
                 kind=bool,
                 default=False,
@@ -97,12 +103,6 @@ class Program(object):
             Argument(
                 names=('collection', 'c'),
                 help="Specify collection name to load."
-            ),
-            Argument(
-                names=('list', 'l'),
-                kind=bool,
-                default=False,
-                help="List available tasks."
             ),
             Argument(
                 names=('no-dedupe',),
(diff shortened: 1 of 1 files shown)


bitprophet:
Thanks for the catch. I assume I chopped up the core-args vs non-core-args stuff that way on purpose, so need to doublecheck the tickets & git history for it to be sure this won't cause another problem. But could've just been an oversight.

jordoncm:
The other solution to this is to not assume `self.args.list` is present here (possibly elsewhere too): https://github.com/pyinvoke/invoke/blob/master/invoke/program.py#L341

bitprophet:
> I assume I chopped up the core-args vs non-core-args stuff that way on purpose

Revisiting this now that it's biting me in my dayjob use - I can see why I initially thought "don't expose `--list`, we're pretending to be a Real Program with Real Subcommands". However, I myself attempted to do `custombinary --list` recently, implying it's intuitive if you're used to Invoke itself.

Given that I hate unintuitive things and it's not super likely that *non*-Invoke users will be bothered by its existence...and of course given that having it in the non-core args is causing problems...I think probably OK to shuffle it around for now, as per this PR.

bitprophet:
Also, @sophacles made the same argument re: inclusion of list in [this comment](https://github.com/pyinvoke/invoke/issues/276#issuecomment-143462511) some time ago (re #276). So one less thing for that branch to handle whenever I get back to it (sob).

bitprophet:
We don't seem to cover the core bug this fixes in either the unit or integration test suites, sadface. Will whip one up.

bitprophet:
All set + tests + changelog, thanks again!

Issue #289, opened 2015-12-18
Prompt/etc autoresponder


bitprophet:
This represents porting Fabric 1.x's limited ability to autorespond to sudo prompts, to Invoke, and making it Better™.

For now, I've gone with a basic "when you see X, always respond with Y" functionality that can be easily configured via a `{pattern: response}` dict.

This isn't the most powerful implementation; e.g. one might want "only respond to patterns once", "respond to X pattern the first 3 times only", "respond to pattern X but only if you've already responded to pattern Y", etc etc.

However it's arguably more flexible than using e.g. `pexpect` (which requires you to anticipate exactly what prompts will happen, in what order; Invoke/Fabric use cases tend to have prompts appear arbitrarily) and I expect it will be enhanced in the future, perhaps by ripping out the responder aspects into their own class or something.

The functionality is already pretty much complete after a ~week of hacking, fucking up, getting confused, becoming less confused, and more hacking. I'm making this ticket just so I have something to link in the changelog & a checklist for the random minor tasks left to perform.

----

Left to do:

* [x] Deal with recently-discovered funkiness in interactive stdin use case
* [x] Make sure the basic functionality works for Fabric 2 as well as with local/subprocess
* [x] Reinstate the Windows support for `handle_input`, which mirrors real stdin to the subprocess' stdin.
* [x] Doublecheck whether we actually need Windows `getch` as per https://github.com/pyinvoke/invoke/pull/304#issuecomment-174005880
* [x] Figure out whether to retain the `sleep` call in `handle_input`; git-blame the Fab 1 line and doublecheck the rationale. Probably do want it back tho, probably chews more CPU than it really needs to as-is.
* [x] Reinstate the "print read stdin data to stdout when pty=False" behavior from fab 1 - it is necessary
* [x] Ditto the "halt stdin mirroring while the local codebase is explicitly trying to prompt for shit it needs" functionality from Fab 1.
    * ...does this mean Fab 1 literally can't deal with users manually executing things like `raw_input` and `getpass`? Cuz if it *can* then that implies said halting/locking might not have been necessary? Or is a corner case?
    * See notes in desc of #294 - probably not going to go down the same road as in Fab 1 which means we wouldn't have to worry about this wrinkle.
* [x] Centralize & normalize the encoding related noise, as per comments by @pfmoore in comments below.
    * Quite possibly via work/discussion being done over on #274 and its linked tickets
[...]

PR #304, opened 2016-01-14
Handle stdin on Windows using msvcrt functions

pfmoore:


bitprophet:
I think I already have this in my #289 branch, but leaving this open in case I'm misremembering or if my implementation is missing something from here.

pfmoore:
Cool - when's that likely to be released? Right now, invoke simply doesn't work on Windows unless I pin to an older version...

bitprophet:
It's at the top of the priority list right now :)

bitprophet:
OK I think my implementation is close enough to yours, given that I refactored where some of these bits live / abstracted them slightly. I'm working on the branch today but pushed its current state: https://github.com/pyinvoke/invoke/compare/rest-of-289

Will close this now but feel free to comment on #289 if you identify more problems with how I'm using `msvcrt`. BTW, I've got some of your other recently opened issues in my milestone too and will be checking on them before I merge. Chances are they'll also want to roll into #289 to limit the number of places we're having discussions...will see :)

pfmoore:
Fantastic - I'll have a look at #289 for you, probably tomorrow. Sorry if I was being a bit pushy over this :-)

pfmoore:
@bitprophet I'm not sure how to comment on that compare link you posted, so I'm going to post here.

Why do you need to use ```msvcrt.getch()``` in ```read_char```? In my testing, ```input_.read(1)``` worked fine. Problems with getch:

* No echo
* Ctrl-C processing is suppressed, there's no way to break out
* Function keys return multi-byte sequences, one byte for each ```getch()``` call, and it's not clear how that interacts with ```kbhit```.

You may get better results with ```getwche```, but honestly I'd stick with ```read()``` unless you know there's a problem with it (I promise I'll let you know ::-))

I assume from the name, that ```read_byte``` *must* only be called with streams opened in binary mode? So not ```sys.stdin``` on Python 3? Otherwise you will get a (Unicode) character. And yet in ```handle_stdin``` you do ```self.write_stdin(self.encode(byte))``` - which doesn't make sense as you don't encode bytes, you decode them - so it looks like you expect ```read_byte``` to return a Unicode character, not a byte.

Overall, it feels to me like your separation of byte and character is muddled. You may be able to get things working for now, but I'd *strongly* recommend reviewing the code and making a clean separation between bytes and characters, or you'll spend forever fighting {de,en}coding issues. I'd be willing to do such a review, if you want - but it's not really appropriate for this change right now. Just avoid ```msvcrt.getch``` and stick with ```read()``` and you should be OK for the moment.

bitprophet:
That code is (aside from the reorganization) ported straight from Fabric 1, whose Windows support (like this project's) comes entirely from users like you. So, I have no idea why `getch()` was considered the right approach at the time :)

I'm totally still not grokking the gory details of Python 3's bytes vs strings nonsense, despite trying to understand it for years, which is why that's all very muddled. I'm not paid enough to get it working perfectly so my MO has been to get things working for the base case on 2 and 3 and to rely on you & others to raise flags when things seriously break (and then to rely on you providing tests to help prove when things work or don't).

Anyway let's take this to #289 from now so I'm not getting 2-3 times the thread notifications ;) thanks!!

PR #330, opened 2016-02-19
Fix tests on Windows

pfmoore:


pfmoore:
Fixes a breakage introduced recently. See https://github.com/pyinvoke/invoke/commit/775e159d2df39a30e4b1761476409c7ee5bd8a9d#commitcomment-16208818 for details.

bitprophet:
> Looks like this changeset broke the Windows tests. It's patching invoke.platform.tty which doesn't exist on Windows. The test is skipped on Windows (below) so this might be easy enough to fix (add tty = None and termios = None to the platform.py module when we're under Windows). I've just submitted PR #330 as a fix.

Derp, sorry. I may have been hoping `@skip_if_windows` would somehow pre-empty the `@patch`; I think simply reversing the decorators so `skip_if_windows` comes first will suffice.

EDIT: yea, it seems to, if I trigger exceptions by telling a `@patch` call to patch a nonexistent-on-unix identifier, then make `@skip_on_windows` skip always (i.e. pretending I'm on Windows), then move them above the `@patch` decorators, things are happy.

Appveyor already ran my branch on 2.7 and it seems to have worked there as well: https://ci.appveyor.com/project/bitprophet/invoke/build/1.0.234/job/ah0xgo0x3h77bh50

> BTW, I only spotted this issue because I updated my fork of the project. Are the main project (master branch) appveyor results publicly visible anywhere?

https://ci.appveyor.com/project/bitprophet/invoke/history is what my browser history says I use, and it works for me even when I'm not authenticated on AppVeyor.

Not sure why they namespace on the github-auth account name instead of the org the repo comes out of, but it's an understandable enough oversight.

bitprophet:
Merged my version of these changes (0f90fa5f6e35a74ff5cf364111a1663e491ba860) to master & just pushed. Gonna close this for now; but this is probably the best place to comment if you find something off with that patchset on your end, @pfmoore - thanks as usual.

PR #332, opened 2016-02-22
escape error log in command failure diagnostics

jhermann:
This prevents Unicode appearing in the error log of failed commands throwing encoding exceptions (thus obscuring the actual error), by using `repr()` to enforce pure ASCII.

Maybe not the best solution, but a robust one.

### DIFF
--- a/invoke/exceptions.py
+++ b/invoke/exceptions.py
@@ -35,6 +35,7 @@ class Failure(Exception):
         if self.result.pty:
             err_label = "Stdout (pty=True; no stderr possible)"
             err_text = self.result.stdout
+        err_text = '\n'.join([repr(i)[1:-1] for i in err_text.splitlines()])
         return """Command execution failure!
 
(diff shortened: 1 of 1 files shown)


bitprophet:
Thanks for this! Can you provide some concrete examples of Unicode-bearing command output that triggers this, plus the stacktrace from before this is applied & then the one from after it's applied? Just want some hard details in front of me so I can make sure I grok it.

I get the basic need & have run into the "this is not my beautiful error" masking problem while poking #350, I just want to make sure I get what your solution is doing & make sure I can't come up with a slightly less baroque one :)

Offhand I'm assuming this is basically saying "give me every byte/char from every line of the output [then join them back etc]" as a way of cutting up any multibyte Unicode nasties - meaning the result could look pretty garbled in some situations (but, not causing errors, so...). But I'm not at all sure, thus the request for clarification.

PR #342, opened 2016-03-25
make Collection subclassable

lsaffre:
I would like to extend the invoke.Collection class by subclassing it. But the from_module method in the current version will always return the original class. This little patch has no influence on the API, is cleaner, and would make me happy :-)

### DIFF
--- a/invoke/collection.py
+++ b/invoke/collection.py
@@ -120,7 +120,7 @@ class Collection(object):
         return self.name == other.name and self.tasks == other.tasks
 
     @classmethod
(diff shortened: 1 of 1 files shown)


bitprophet:
Thanks, that's an oversight/bug for sure, adding to milestone :)

bitprophet:
Years later...sob...merged this & made sure it played nice with how the code changed in the interim. (tl;dr we moved things around & are using an inner closure, but were still hardcoding `Collection` instead of `cls`.)

lsaffre:
Great! Thanks!

PR #356, opened 2016-06-04
Remove non contextualized tasks

KuangEleven:
Issue #114 Removes non contextualized tasks from the code as well as the ctasks convenience decorator. Docs are updated to match.


MrAlias:
This appears to have introduced backwards incompatible changes with out a major version change.

The main one for us being [this](https://github.com/pyinvoke/invoke/pull/356/files#diff-a2c23243533cad12946af2f1122bb4ebL315).

We have adapted to fix the issue on our end, but this should likely be addressed to avoid future issues with other projects.

PR #363, opened 2016-06-21
add missing ctx parameter

moonshadow:


indera:
👍

bitprophet:
Hah, we missed a spot. Thanks!

Issue #384, opened 2016-08-23
Config loading and locale bug


PatrickMassot:
The way [config.py](https://github.com/pyinvoke/invoke/blob/322718d7f38ce04fc2bde947ba67ab4002f669b6/invoke/config.py#L499) checks for existence of config files works only for English speakers. Try the following after installing the relevant locale:

``` python
import locale
from invoke import task

locale.setlocale(locale.LC_ALL, 'fr_FR.utf8')

@task
def test(ctx):
    pass
```

Then `inv test` crashes with `FileNotFoundError: [Errno 2] Aucun fichier ou dossier de ce type: '/etc/invoke.yaml'` and its French error message.

indera:
I think you have used an invalid locale name.
Please try to use locale name: 

locale.setlocale(locale.LC_ALL, '**fr_FR.UTF-8'**)

To view a list of valid locales:

```
$ locale -a | grep fr
fr_BE
fr_BE.ISO8859-1
fr_BE.ISO8859-15
fr_BE.UTF-8
fr_CA
fr_CA.ISO8859-1
fr_CA.ISO8859-15
fr_CA.UTF-8
fr_CH
fr_CH.ISO8859-1
fr_CH.ISO8859-15
fr_CH.UTF-8
fr_FR
fr_FR.ISO8859-1
fr_FR.ISO8859-15
fr_FR.UTF-8
```

PatrickMassot:
Thank you for your comment. I think I would have noticed if setting locale this way didn't work in all my python programs. Let's see:

```shell
$ locale -a | grep fr
fr_BE.utf8
fr_CA.utf8
fr_CH.utf8
fr_FR.utf8
fr_LU.utf8
```

Interesting but irrelevant to invoke I'm afraid.

indera:
@PatrickMassot did you try any other locale such as "en_US.utf8"?

Please include the full trace message before the "FileNotFoundError" you have encountered.

PatrickMassot:
Full trace is 
```
Traceback (most recent call last):
  File "/usr/local/bin/inv", line 11, in <module>
    sys.exit(program.run())
  File "/usr/local/lib/python3.5/dist-packages/invoke/program.py", line 270, in run
    self.execute()
  File "/usr/local/lib/python3.5/dist-packages/invoke/program.py", line 380, in execute
    executor = self.executor_class(self.collection, self.config, self.core)
[...]

Issue #364, opened 2016-06-21
Consider dropping/deprecating Python 2.6 support


bitprophet:
## Background

There are a number of arguments among progressive package maintainers for dropping 2.6:
- Not supported by the Python core team (including security bugs)
- Increasing numbers of third-party packages are also dropping support or at least doing this same sort of side-eye at the possibility.
  - Including pip, which isn't really "third party", and tends to drag the rest of us screaming along with it
  - Sphinx, as of late 2016, has also dropped Python 2.6 support
- Debian 7 and up include 2.7; RHEL/CentOS 7 too
  - Users still stuck on RHEL5 must be backporting their own Python if they're on 2.6 (as system default was 2.4) so they arguably can just as easily get 2.7
- Usual "backporting packages is a core need for good systems administration, as is not using unsupported software" argument
- (Admittedly minor, IMHO) syntax niceties like "you can use empty brace expressions in strings, dict comprehensions, etc"

But there are strong counter-arguments, which is why we currently still support 2.6:
- Through recently, a nontrivial number of users (across my personal maintained packages) are still on 2.6 - as high as 25-33% for Paramiko for example.
  - EDIT: though we just re-ran the numbers and it's now at only 6%! 60k downloads over 2 weeks, compared to 873k in same period for 2.7 (plus another ~45k for Python 3). Fewer, but still not trivial. Fabric's at 8.5%.
  - We're not sure _why_ so many users are still on 2.6...
  - Though it could simply be "it's RHEL/CentOS 6", which includes Python 2.6 as the default
    - Presumably has a very large install base (users of these distros trend conservative) given 7 only came out in 2014
- Supporting 2.6 (in a Python 2+3 world) isn't, technically speaking, significantly more difficult than supporting 2.7.
- "You're doing it wrong" (at folks who could potentially move to 2.7, but haven't yet) always feels like a shitty argument to use to justify making maintainers' lives easier, _when_ the thing under discussion is comparatively easy to keep doing.
  - The key here is that it's becoming rapidly less easy, due to many other packages dropping support...
## Possible way forward
- Add a `DeprecationWarning` (as does pip, cryptography, etc) stating that Invoke 1.0 (maybe 2.0, which is likely to not be years after 1.0) will drop Python 2.6 support. Include a link to this ticket & "please explain why you're still using 2.6".
- Wait a while, see if any other good reasons pop out of that, that aren't already in the counter-arguments list.
[...]

PR #385, opened 2016-08-23
Catch FileNotFoundError in config.py - issue #384

indera:


PatrickMassot:
It's very nice to fix #384 but why did you drop python 2.6 in the same PR??

Have you seen #364?

bitprophet:
Thanks for this (& raising the issue itself in #384 - sadly I don't have the time nor expertise to have full international support in my initial code :D). Some feedback:

- As @PatrickMassot notes we still support 2.6 pending #364, so please make sure the PR works there :)
- What's `.ropeproject`? I generally prefer people use their own gitignores and not the project's gitignore, for anything the project itself doesn't require for development. So dropping that part of the patch would be appreciated too.
- I'm slightly torn on the actual, core fix: obviously introspecting the error message for English is non-ideal, but I worry expanding the scope to simply "any `OSError` or `IOError`" could hide other subtypes of such errors.
    - Part of me wonders if it's worth testing specifically for `FileNotFound` on Python 3; but the syntactical overhead of doing that probably isn't worth it (until/when we drop Python 2).
    - Overall, though, the chance of some error at this point that is meaningfully different from "file not found" feels low, and users in those situations who see behavior they're not expecting, will hopefully enable logging. So yea. I'll probably merge this :)

indera:
@bitprophet .ropeproject/ folder is generated by a vim python plugin I am using: https://github.com/klen/python-mode

indera:
Fails on windows but other than that looks good to me

Issue #313, opened 2016-01-23
@task seems to eat docstring


thebjorn:
I'm trying to document my task library using Sphinx, and ran into a problem with docstrings and autodoc.  How to reproduce (before/after functions added to verify that sphinx works):

```
(sphinxtst) go|c:\srv\tmp\ctaskdocs> cat tasklib.py
from invoke import task, ctask, run

def before():
    "before docstring"
    pass

@ctask
def hello_ctask(ctx):
    "hello ctask"
    ctx.run('echo world')

@task
def hello_task():
    "hello task"
    run('echo world')

def after():
    "after docstring"
    pass
```

A minimal conf.py

```
(sphinxtst) go|c:\srv\tmp\ctaskdocs> cat docs\conf.py
# -*- coding: utf-8 -*-
import sys, os
extensions = ['sphinx.ext.autodoc']
```

and a minimal root document using autodoc

```
(sphinxtst) go|c:\srv\tmp\ctaskdocs> cat docs\contents.rst

tasklib
-------

.. automodule:: tasklib
   :members:
   :undoc-members:
```

Set the pythonpath to directory containing tasklib.py to keep sphinx happy

```
(sphinxtst) go|c:\srv\tmp\ctaskdocs> set PYTHONPATH=%CD%

(sphinxtst) go|c:\srv\tmp\ctaskdocs> sphinx-build docs build
```

The resulting build/contents.html only contain references to the functions that do not have the `@task/@ctask` decorators.

thebjorn:
Well.. after some meditation, a bit of food, and re-reading the sphinx docs, I've come up with an incantation that works:

```
tasklib
-------

.. automodule:: tasklib
   :members:

.. autoclass:: tasklib.hello_ctask
.. autoclass:: tasklib.hello_task
```

ie. remove `:undoc-members:` on automodule, and add an autoclass directive for every function that has a `@task/@ctask` decorator.  Neither `:members:` not `:undoc-members:` should be specified on autoclass (or you'll get all of the Task class' innards).

The various warnings about `"no attribute '__bases__'"` and duplicate object description (between the Task class and the Task instance) should all be benign.

thebjorn:
The ``autoclass`` approach turned out to be sub-optimal since it just included the task's docstring, and not the task's parameter list (or parameter help).  Maybe a solution would be to inject the parameter list into the task's docstring, but that seems very hackish.

[...]

PR #399, opened 2016-10-20
Added support for decorated tasks

frol:
Currently, there is no way to use decorators with Invoke tasks:

``` python
import functools

def my_decorator(func):
    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        print("Decorated call")
        return func(*args, **kwargs)
    return wrapper

@task
@my_decorator
def my_task(context, name, option1=1):
    print("Hello, %s!" % name)
```

This code will complain about missing `context` positional argument due to the fact that the `wrapper` defines `*args, **kwargs`.

Adding this small patch, `argspec` will respect `__wrapped__` field pointing to the decorated function.

### DIFF
--- a/invoke/tasks.py
+++ b/invoke/tasks.py
@@ -132,6 +132,9 @@ class Task(object):
         # TODO: __call__ exhibits the 'self' arg; do we manually nix 1st result
         # in argspec, or is there a way to get the "really callable" spec?
         func = body if isinstance(body, types.FunctionType) else body.__call__
+        # Handle decorated task functions
+        if hasattr(body, '__wrapped__'):
+            return self.argspec(body.__wrapped__)
         spec = inspect.getargspec(func)
         arg_names = spec.args[:]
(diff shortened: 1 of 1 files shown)


bitprophet:
Possible dupe of #246, and linking to possibly related #313

frol:
This PR doesn't address #313. However, it is indeed covers the #246, though I find my implementation much simpler.

frol:
So?...

bitprophet:
Please don't bump tickets, it's rude & wastes the maintainers' time :(

frol:
Please, don't ignore at least pull requests (i.e. others time to read your code and make a patch), it is rude and doesn't make any good for a project.

_You haven't got back to #189, #190 (just a few examples of PRs) with **any** response at all after **TWO YEARS**, so I will bump the tickets unless I give up on this unmaintainable heap of buggy code_.

bitprophet:
I seriously don't have time for that sort of nonsense. Blocked :(

Issue #33, opened 2012-10-21
Task-list format option


bitprophet:
I.e. instead of just `inv --list` spitting out the current default flat list, allow things like `inv --list nested` for a nested/indented display, `inv --list=foo` to just list the tasks within namespace `foo`, etc ad infinitum.

LATER EDITS: so a lot of strongly related tickets have popped up and there seem to be two major currents:

- Formatting, i.e. flat with dots vs nested vs machine-readable
- Content, i.e. all tasks vs just namespaces vs just tasks within a given namespace

Seems unwise/impossible to shoehorn all of that under a single optional value to `--list` so I'm sure we'll grow some additional flags either way, with the main question being which (if any) such option deserves to live as the optional `--list`value.

Brainstorm:
- Format could be a single flag with string option, e.g. `--list-format (flat|nested|machine)` (what term do other tools use for "machine-readable" anyways? Can't remember offhand...LATER EDIT: actually these days we often just use JSON for that and just call it eg `--list-format=json`) or it could be a handful of booleans like `--list-nested`, `--list-machine`.
  - Former feels easier to implement, harder to fuck up by giving >1 at a time, and easier for users to grasp (i.e. fewer options to think about) so it seems like a no-brainer.
- Content needs levers for 'root' (though this needs to not clash with the related-but-distinct `--root` flag which changes where tasks files are sought from; but perhaps we should be, well, _namespacing_ those as well so they are renamed to eg `--discovery-root`) and 'depth` and I think that's it, since those together can enable just about anything:
  - "Show me everything": root is the top of the namespace, depth is None
  - "Show me just the namespaces": root is top of namespace, depth is 1
  - "Show me just tasks under namespace foo"
- Might actually need another lever for the confluence of those two, i.e. "I am listing some subset of tasks and I want to tweak whether they are shown as full paths or sub-paths"?
  - Though at *this* point I wonder if users who want a non-full-path display (i.e. `inv --namespace foo --list` spitting out task names like `bar` and `baz` instead of `foo.bar` and `foo.baz`) should be relying on "each Python module can be its own standalone namespace" and doing things like `inv --collection=namespace --list"` instead (see [Collection loading](http://docs.pyinvoke.org/en/latest/concepts/loading.html#configuring-the-loading-process).)
    - Since, if one truly wants to 'focus' on a given sub-namespace, giving it to `--collection` automatically ensures that *everything* - display, invocation, etc - instantly works as you'd expect.
[...]

PR #408, opened 2016-11-22
Change task list sorting

samfoo:
Grouping by parent, and then sorting by depth/lexically makes the output dramatically more readable, this commit changes task lists sort to be:

1. By max hierarchy depth
2. Grouped by parent hierarchy name
3. Lexically

This (mostly) preserves the previous behaviour/requirement of top level tasks being more important, thus shorter, thus closer to the top of the list. Grouping tasks by their parent, however, makes deeply nested hierarchies easier to find/read. 

Fixes pyinvoke/invoke#402

This splits sorting for aliases (they're still sorted by depth, then lexically) from sorting for task names.

### DIFF
--- a/invoke/config.py
+++ b/invoke/config.py
@@ -27,7 +27,7 @@ else:
 
 from .env import Environment
 from .exceptions import UnknownFileType
-from .util import debug, merge_dicts, AmbiguousMergeError
+from .util import debug, merge_dicts
 from .platform import WINDOWS
 
 
@@ -594,5 +594,3 @@ class Config(DataProxy):
                 continue
             data[key] = value
         return data
-
-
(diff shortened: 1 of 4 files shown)


samfoo:
Hmmm... the patch code coverage failure is just from moving the `merge_dict` method to `util.py`. I can add tests, but it is a bit orthogonal to the rest of the changes here.

bitprophet:
Don't remember how strongly I cared about having CodeCov run the patch-specific coverage check; offhand I think I care much more about the project overall, which is green here. So, I'll look at that briefly but not worried. Thanks for this! Adding to milestone.

bitprophet:
Grumble, this got overlooked when I recently overhauled task list formatting stuff (#33). Gonna close this as hopefully now duplicative but please check out the results of that ticket (out since 0.23 and we're now into 1.0.x) and drop a line if there's still use cases unserved. Thanks!

Issue #204, opened 2015-01-11
Bundled libs in invoke/vendor


athmane:
Hello,

I'm packaging  pyinvoke for Fedora (preparing for Fabric 2.x), and the policy does not allow bundled libraries, so I'm wondering if we could remove the libraries in ./invoke/vendor and replace them with disto/pypi provided one.

Note that Debian has the same policy, so this might help Debian packager too.

Here's the patch against 0.9.0 branch: https://athmane.fedorapeople.org/pkgs/invoke-0.9.0-remove-bundled-libs.patch

Thanks.

bitprophet:
The rationale for vendorizing is here: http://bitprophet.org/blog/2012/06/07/on-vendorizing/

It's far from a perfect solution and I've definitely second-guessed it at times, but for now the benefits still outweigh the costs.

I'm open to massaging things to make flipping between vendor/no-vendor easier, e.g. tweaking `sys.path` in the top level `__init__` such that we don't need to refer to `vendor.*` - though this feels fraught with peril.

Other ideas welcome too. Most of what I can think of is either kinda gross like the above, or not feasible for what I think your needs are (e.g. we could honor an env var at import time and conditionally import from vendor or not - but then that requires all users of your distributed version to have this env var set, which is probably difficult to manage.)

I know there are other Python libs which aren't budging on the vendor issue; is distro policy for them to just move the burden of un-vendorizing into patches like the one you linked?

tl;dr I don't want to be difficult for OS vendors but I also do have reasons to vendor =/

bitprophet:
Referencing #4 as well, though most of the articulated arguments are in that blog post.

athmane:
I'm in favor of simplifying distro packager's patches by making flipping between vendor/no-vendor easier, that reduce the chance of rebasing the patch very often.

Let me know if you need help with this.

Thanks in advance.

bitprophet:
Implemented in #412 !

PR #412, opened 2016-12-11
Change vendor imports so that it will use system wide modules if the the import from vendor directory fails.

athmane:
This will help maintaining pyinvoke package in most distro that does not accept bundled libraries (in my case Fedora).

The plan is to clear vendor/ and let pyinvoke use system provided modules.

### DIFF
--- a/invoke/collection.py
+++ b/invoke/collection.py
@@ -1,8 +1,14 @@
 import copy
 import types
 
-from .vendor import six
-from .vendor.lexicon import Lexicon
+try:
+    from .vendor import six
+    from .vendor.lexicon import Lexicon
+except ImportError:
+    # Use system wide modules
+    import six
+    from lexicon import Lexicon
+
 
 from .config import merge_dicts
 from .parser import Context as ParserContext
--- a/invoke/config.py
+++ b/invoke/config.py
@@ -3,11 +3,19 @@ import json
 import os
 from os.path import join, splitext, expanduser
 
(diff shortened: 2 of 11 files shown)


bitprophet:
Follow-up on #204 I take it? :D thanks!

This is an interesting reversal of my earlier assumption from that ticket, re: what a common solution might have been - namely to prefer global when present, with vendored as a fallback. In this case, you're preferring vendored, falling back to global.

I think that's good, because it means the only time the vendor copies will be skipped is when the installer/packager is explicitly choosing to forego the vendoring (& is thus implicitly responsible for ensuring dependencies are compatible versions).

Just curious, are you planning to pair this with a build script / `.spec` / whatever, which just does `rm -rf invoke/vendor && (pip|yum|dnf|etc) install <all the deps>`?

athmane:
> Just curious, are you planning to pair this with a build script / .spec / whatever, which just does rm -rf invoke/vendor && (pip|yum|dnf|etc) install <all the deps>?

Yes (close enough), the only issue I think of is if the vedorized library is forked from upstream, in that case the changes should be merged upstream or downstream package.

bitprophet:
Of course. IIRC my vendored copies are untouched aside from occasional hacks required to _be_ vendorized, or (in the case of fluidity, though this was years ago so I'd hope it's released now) unreleased-but-public commits.

If you identify spots where that isn't the case, certainly let me know!

PR #430, opened 2017-02-18
Fix yaml import since only vendor-ized variant has version number suffix

athmane:
Related to #412

### DIFF
--- a/invoke/config.py
+++ b/invoke/config.py
@@ -13,10 +13,7 @@ try:
 except ImportError:
     # Use system modules
     import six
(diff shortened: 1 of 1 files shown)


bitprophet:
Huh, wonder if whomever updated the imports to fallback to external dependencies actually tested it. Yea, IIRC PyYAML does install as just `yaml` when installed regularly. Thanks for the catch!

athmane:
That was me :), I tested on an unclean environment (laziness I guess).

Issue #390, opened 2016-09-15
Tasks modules are not found on sys.path


rogual:
[Docs](http://docs.pyinvoke.org/en/latest/concepts/loading.html#task-module-discovery) say:

> First, if a valid tasks module by that name already exists on Python’s sys.path, no more searching is done – that module is selected.

This doesn't happen, and, looking at the code for `Loader` and `FilesystemLoader`, it doesn't look like there is any code trying to accomplish this.

bitprophet:
The `imp` module loads from `sys.path` (it's the implementation of Python's builtin `import` mechanics). See https://github.com/pyinvoke/invoke/blob/9e1e85a68902c4e2fc481e7d39884c89ab93c2c3/invoke/loader.py#L53

If you can post more details about your tasks file not loading (filesystem layout, contents of module, output of `invoke --list` with `INVOKE_DEBUG=1` set in the shell env, etc) maybe we'll spot a clue. Thanks!

rogual:
Here's a complete example using an empty module:
```
~ $ mkdir abcde

~
$ touch abcde/mything.py

~
$ INVOKE_DEBUG=1 PYTHONPATH=abcde invoke -c mything
invoke.program._parse: argv given to Program.run: None
invoke.program.normalize_argv: argv was None; using sys.argv: ['/Users/me/.pyenv/versions/2.7/bin/invoke', '-c', 'mything']
invoke.program.parse_core_args: Parsing initial context (core args)
invoke.parser.__init__: Initialized with context: <parser/Context: {'hide': <Argument: hide>, 'help': <Argument: help (h)>, 'config': <Argument: config (f)>, 'write-pyc': <Argument: write-pyc>, 'list': <Argument: list (l)>, 'collection': <Argument: collection (c)>, 'echo': <Argument: echo (e)>, 'version': <Argument: version (V)>, 'warn-only': <Argument: warn-only (w)>, 'debug': <Argument: debug (d)>, 'pty': <Argument: pty (p)>, 'root': <Argument: root (r)>, 'no-dedupe': <Argument: no-dedupe>, 'complete': <Argument: complete>}>
invoke.parser.__init__: Available contexts: {}
invoke.parser.complete_context: Wrapping up context None
invoke.parser.parse_argv: Starting argv: ['-c', 'mything']
invoke.parser.handle: Handling token: '-c'
invoke.parser.handle: Saw flag '-c'
[...]

PR #431, opened 2017-02-26
bug #390 fallback to importing modules from sys.path

lmiphay:
This patch adds a fall back to searching sys.path for modules, which I think makes sense (allow a projects modules to be found before looking in sys.path).

However the documentation clearly says the opposite (First... Failing that...).

To follow the documentation, switch the search_path assignment to: search_path =  sys.path + parents

imp.find_module won't search sys.path by itself when an explicit path (second arg) is passed to it.


bitprophet:
Linking to #390

Issue #344, opened 2016-04-09
Allow limiting stdout/err capture


bitprophet:
This is Invoke's version of fabric/fabric#800.

Unlike the notes at the bottom of that ticket, where we're using Fabric 1's `RingBuffer`, here in Invoke land we can just use `collections.deque(xxx, maxlen=yyy)`.

Also, because we have a somewhat richer (and publicly documented) `Result` class, we can (and should) add a flag to `Result` noting when the ring buffer is being limited, so users have at least a little warning when shooting themselves in the foot ("hey, I turned on this memory saving feature I saw on StackOverflow, but now I only see the last 1024 bytes of stdout!")

bitprophet:
While rereading this after receiving #451, I had a realization that did not occur to me at time of posting - namely, are there any _downsides_ to switching to a deque, for the common case of not limiting buffer size? I.e. what do deques sacrifice vs lists?

Going by [the 2.6 docs](https://docs.python.org/2.6/library/collections.html#collections.deque), the only downside _seems_ to be that random access in the middle of the structure trends towards O(n). Clearly not an issue for this use case. But wanted to get this on record anyways.

bitprophet:
Note: this needs care when intersecting with the watchers functionality, since that API is "we give you the current, full copy of the capture buffer on each tick".

At the very least we just need a warning block reminding folks that if they configure a limited buffer, they will necessarily only get that buffer in their watchers, and will need to manually reconstruct the entire stream if that's what they really need. (Though this seems...to defeat the point...)

fruch:
maybe just a way to indicate where is the new data. in most cases I've seen those care about what is new data, for example echo it into logger, looking for specific lines/strings.

PR #451, opened 2017-06-03
#344 allow limiting stdout/err capture

lmiphay:
This PR should hit the _specific todo_ list of #344.

It addresses an issue when a long running invoke thrashes after hitting multiple GB of VSS/RSS.


bitprophet:
Nice, thanks for this! Adding a note to #344 that occurred me to as I re-read it, but I don't believe it's a blocker, so I'll try to merge this soon.

PR #373, opened 2016-07-06
Use inspect.signature to enable Python 3 only features when defining tasks

rsxm:
Fix  #372

Issue #372, opened 2016-07-06
Allow invoke to run tasks with type hinting


jr-minnaar:
Issue #248 mentions type hinting, but I think that is for Invoke to use type hinting internally. This issue is to allow Invoke to run tasks with type hinting annotations.

With current master, when defining a task like this:

```
@task
def hello(name: str = 'Anonymous'):
    print('Hello, {name}!'.format(name=name))
```

I get the following error message when trying to run this task:

> ValueError: Function has keyword-only arguments or annotations, use getfullargspec() API which can support them

I followed the advice and replaced getargspec() in tasks.py with getfullargspec() and everything seems to work.

I don''t know the full implications of this change, but I'll create a fork and run the tests and see if something breaks.

bitprophet:
See #357 :)

jr-minnaar:
Thank you! Don't know how I missed that one, I thought I was thorough with my search of existing issues :-)

FranklinChen:
I'm sad to see this wasn't fixed for the 1.0.0 release that just came out.

Issue #357, opened 2016-06-04
Using python3 keyword-only arguments or annotated arguments causes a ValueError from inspect


KuangEleven:
Defining a task that uses the new Python3 keyword-only or annotation argument syntax and invoking causes a `ValueError: Function has keyword-only arguments or annotations, use getfullargspec() API which can support them` and halts.

Per the docstring for Python3 inspect, getfullargspec() could be used, but that function is itself deprecated in favor of inspect.signature() Unfortunately, neither of those functions exists in Python 2.6, so some level of cascading will be necessary.

For example, neither of these can be invoked:

``` python
from invoke import task

@task
def func(ctx, *args, keyonlyarg=42):
    print(keyonlyarg)

@task
def func2(ctx, x:int=3):
    print(x)
```

PositiveAlex:
It is also not woking with annotations.

```
@task
def test(context: object, suite: str = None, filter_tag: str = None):
```

FranklinChen:
Any word on this? I use Python 3 with `mypy` for types and just realized that I'm blocked on using `invoke` right now.

bitprophet:
Thanks for the bump, I'll take a look at #373 soon which seems to be the main patch around this.

@FranklinChen If you can check out that PR's branch and confirm it works against latest master & solves your particular flavor of the issue, that'd be a big help!

FranklinChen:
@bitprophet Yes, I am actually using the fork at #373 and it works for me.

nathanielford:
Hi! I'm still seeing this with version `1.1.1`. 
```
nford@Nathaniels-MacBook-Pro:~/repos/deploy-apps $ invoke --list -v
Traceback (most recent call last):
  File "/usr/local/bin/invoke", line 11, in <module>
    sys.exit(program.run())
  File "/usr/local/lib/python3.6/site-packages/invoke/program.py", line 324, in run
    self.parse_collection()
  File "/usr/local/lib/python3.6/site-packages/invoke/program.py", line 408, in parse_collection
    self.load_collection()
  File "/usr/local/lib/python3.6/site-packages/invoke/program.py", line 599, in load_collection
    module, parent = loader.load(coll_name)
  File "/usr/local/lib/python3.6/site-packages/invoke/loader.py", line 76, in load
    module = imp.load_module(name, fd, path, desc)
[...]

Issue #248, opened 2015-05-17
Considered using PEP-484 type annotations?


Dedoig:
They will be the standard way to type annotate function signatures (including parameterized types like lists, dicts, etc.), and are run-time accessible. Is it worth looking into using that?

bitprophet:
Thanks for asking!

I can't remember offhand whether there are existing plans for backporting that feature to Python 2. My default instinct is to avoid Python 3 specific things (even in an optional mode) because they generally don't add enough benefit to be worth the expected support burden (i.e. confused users trying to use those features on Python 2; extra code divergence between the interpreters; etc).

So, I'm neutral on the overall idea but it hinges on that question. Let me know if you look into that.

tuukkamustonen:
At least mypy (where PEP-484 originates from) lists python 2.x support for 2015 in their roadmap: http://mypy-lang.org/roadmap.html

bitprophet:
Linking this to #357 which is semi related (it's re: folks using this functionality in Python 3, for its own sake, not re: any internal support).

elgalu:
hi! are there plans for https://github.com/python/mypy support?

PR #458, opened 2017-07-10
Use getfullargspec if available

Bouke:
I like using annotations in Python as it allows my IDE to provide completions:
```python
@task
def test(c: Connection):
    c. # < IDE can now offer completions based on the Connection type
```

[`inspect.getargspec`](https://docs.python.org/3/library/inspect.html#inspect.getargspec) is deprecated since Python 3. It is still available on 3.6, however it throws an error when one uses function annotations:

> ValueError: Function has keyword-only parameters or annotations, use getfullargspec() API which can support them

The new api `inspect.getfullargspecs` is a drop-in replacement according to the stdlib authors:

> Deprecated since version 3.0: Use getfullargspec() for an updated API that is usually a drop-in replacement, but also correctly handles function annotations and keyword-only parameters.

This PR tries to use `getfullargspecs` and if that's not available, falls back to `getargspecs`.

### DIFF
--- a/invoke/tasks.py
+++ b/invoke/tasks.py
@@ -132,7 +132,12 @@ class Task(object):
         # TODO: __call__ exhibits the 'self' arg; do we manually nix 1st result
         # in argspec, or is there a way to get the "really callable" spec?
         func = body if isinstance(body, types.FunctionType) else body.__call__
-        spec = inspect.getargspec(func)
+        # getargspec is deprecated on Python 3 and throws an error if 
+        # annotations are used.
+        try:
+            spec = inspect.getfullargspec(func)
+        except AttributeError:
+            spec = inspect.getargspec(func)
         arg_names = spec.args[:]
         matched_args = [reversed(x) for x in [spec.args, spec.defaults or []]]
         spec_dict = dict(zip_longest(*matched_args, fillvalue=NO_DEFAULT))
--- a/invoke/tasks.py
+++ b/invoke/tasks.py
@@ -132,7 +132,7 @@ class Task(object):
         # TODO: __call__ exhibits the 'self' arg; do we manually nix 1st result
         # in argspec, or is there a way to get the "really callable" spec?
         func = body if isinstance(body, types.FunctionType) else body.__call__
-        # getargspec is deprecated on Python 3 and throws an error if 
+        # getargspec is deprecated on Python 3 and throws an error if
         # annotations are used.
         try:
             spec = inspect.getfullargspec(func)


bitprophet:
Thanks! This should be linked to #248 / #373 / #372 and I suspect the PRs in that list may obviate this one (or vice versa). No time to compare at the moment but if someone else does, please comment!

haydenflinner:
FYI the vendored `decorator` includes this fallback as well, IIRC it does something a little more involved but it's been a while since I last looked

neighthan:
It would be nice if we could merge in one of the PRs to make `invoke` work properly on Python 3. As this one is extremely short (code review should take just a minute) and does exactly what the error message says should be done to avoid this error, it seems a good candidate. There are other PRs (well, at least one) that address the same issue, but I think just getting _something_ merged now would be a great benefit. It's been two and a half years since #357 was created, and I'd be glad to stop fixing it myself with each install of invoke.

haydenflinner:
@neighthan rather than patching your installs of invoke, just `pip install magicinvoke`, it does the same thing but includes a few more bugfixes.

nyurik:
@bitprophet hi, thanks for all the hard work on this project!  This PR seems like a simple solution to using typing in Python3 -- are there any blockers for merging it?  Thanks!!!

MichaelRBond:
Any movement on getting this small PR merged in?

Crocmagnon:
I'd love to see this merged too 🙏🏻

neozenith:
Hi 👋 
I'm starting to volunteer on triage on PRs and issues.

The messaging for the roadmap can be found here:

Key items that are putting back pressure are:
 - Migrating CI [DONE]
 - Dropping Python2 support

The latter task is the more important task which will make future type annotation support easier.

Given this PR has a high number of 👍 votes on it I'll take that as community feedback it has good ROI for me to investigate. We do appreciate everyone's patience as we have limited volunteering time and cognitive bandwidth.

neozenith:
This also relates to PR #606

Crocmagnon:
I'd argue on the contrary that #606 duplicated this work, given the timeframe 🙂

manykarim:
With the recent Python 3.11 Release. the `getargspec()` has been removed.
https://docs.python.org/3.11/whatsnew/3.11.html

So as of now, `invoke` does not work in Python 3.11 when using the `Tasks` decorator.
Would be great if this PR could be merged, as it seems to be a minimal change.

Is there anything I could do to support?

Mateusz-Grzelinski:
this fixes https://github.com/pyinvoke/invoke/issues/891

caphrim007:
@bitprophet i can also verify this patch addresses python 3.11. would be super cool if it could be merged.

bitprophet:
2.0.0 will be out soon & is now using `inspect.signature`!

PR #464, opened 2017-08-02
Catch missing file and reflect it internally

Arrowbox:
If a given prefix does not result in any hits (yaml, yml, json, or py) then we should mark _{prefix}_found as False.

Both SourceFileLoader and imp.load_source kick up IOErrors when the file is missing.

Tested against python 2.7 and 3.5 on Linux.

### DIFF
--- a/invoke/config.py
+++ b/invoke/config.py
@@ -12,14 +12,10 @@ if six.PY3:
     except ImportError: # PyPy3
         from importlib._bootstrap import _SourceFileLoader as SourceFileLoader
     def load_source(name, path):
-        if not os.path.exists(path):
-            return {}
         return vars(SourceFileLoader('mod', path).load_module())
 else:
     import imp
     def load_source(name, path):
-        if not os.path.exists(path):
-            return {}
         return vars(imp.load_source('mod', path))
 
 from .env import Environment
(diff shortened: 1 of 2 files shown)


Arrowbox:
Need to handle errors on Windows. It looks like a OSError(winerror = 2) is raised, which is FileNotFound. I'm tempted to just do an `if not os.path.fileexists(path): raise IOError` in the `load_source` function. I guess you could argue that between the os.path.fileexists and SourceFileLoader.load_module calls the file could be deleted.

bitprophet:
Yea, the "normalize into IOError" approach is tempting but you're quite right that it's technically susceptible to a race condition. (More likely perhaps than we guess, I can easily see a situation where e.g. one's running `inv` on a server, and forgetting that it's being config-managed in parallel, and that config management is replacing or nuking the config file...)

So I think an explicit check is probably safer? e.g.: 

``` python
except (IOError, OSerror) as e:
    if isinstance(e, OSError) and hasattr(e, 'winerror') and e.winerror is not 2:
        raise
    # otherwise, do what we do now...
```

Probably possible to artificially construct such an error in the test suite too, so we cover that branch. (I don't usually do anything Windows specific in tests besides skip things, but in this particular case it seems too easy to not do.)

PR #469, opened 2017-08-09
Update library.rst

Aiky30:
Following the instructions it was not obvious how the subclassed config should be added to the program definition. I needed to read the program API to find the relevant information: http://docs.pyinvoke.org/en/latest/api/program.html#module-invoke.program.

All examples previous to this section have been accurate and the examples contain all of the relevant information to get the example working.

### DIFF
--- a/sites/docs/concepts/library.rst
+++ b/sites/docs/concepts/library.rst
@@ -192,8 +192,10 @@ here:
 - Default configuration values - less critical (most defaults aren't labeled
   with anything Invoke-specific) but still sometimes desirable.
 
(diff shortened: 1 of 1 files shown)


bitprophet:
I feel like the "here's specifically how to take your custom Config class and give it to a Program" bit should live in the `note::` itself perhaps. Either way, thanks for the input!

Aiky30:
I agree and started to add it to the note hence the addition of a note segregation. The reason that I opted against it and added it to the example is because the note is outside of the relevant section titled: "Changing filenames and/or env var prefix". It could be possible to just create a note in that section I guess. 

Happy to help more regarding this if required.

Issue #445, opened 2017-05-22
Tasks without context


pekkaklarck:
I haven't really needed context objects in my projects and, as I already commented #114, they seem to mainly complicate writing good APIs. I would really like to have a separate `@task` that works like the standard variant but doesn't require context as the first parameter. On a quick look at the source this ought to be pretty easy to implement and the hardest part would be figuring out a good name for the new decorator. If this isn't considered a totally unwanted idea, I'm willing to provide a PR.

bitprophet:
I plopped some thoughts in the old ticket, https://github.com/pyinvoke/invoke/issues/114#issuecomment-303466757 - adding this to 1.0 milestone as a possibility.

pekkaklarck:
Thanks considering this! I did some hacking and was able to get something working. My approach was adding new `context=True` argument to the `Task` class and using this information to see does a task need a context or not. If this doesn't sound totally stupid, I can push my changes to my fork for others to see them.

Do you have ideas what the new decorator should be called? One option would be making this an optional argument to the current `@task`, but having a separate decorator would be more convenient.

bitprophet:
Yea, the 'argument or distinct decorator' question is open:

- Fabric 2 will grow its own `@task` so multiple types of 'task decorators' would be a thing
- However it's hard to come up with a useful, non verbose name, something used so frequently doesn't really want to be named e.g. `@contextless_task` or `@implicit_task` or etc, IMO.
- Could do something like `@ctask` (this is what we did before), `@itask`, etc, but those sacrifice obviousness for tersity
- Given that the Fabric version of the decorator can rely on Python module namespacing (i.e. it would be `fabric.task` instead of `invoke.task`), which sidesteps the naming problem for that new decorator, I wonder if the argument approach for this use case makes more sense.
  - E.g. it's a parameterization of the "Invoke-level task" concept and not a wholly new type of task?

Other ideas welcome, this is just my sleep deprived brainstorm.

pekkaklarck:
Having a separate task would be convenient. Perhaps it would be best to have explicit names like `contextless_task` or `task_without_context` and let users import them `as better_name` if they want to? I my current context I'd probably do `from invoke import contextless_task as task`.

pekkaklarck:
Here's my branch to implement contextless tasks. Tests and docs totally missing and I'm sure I have forgotten about something else too. Comments about the approach would be appreciated. This this looks OKish, I try to find time to finish the work and create a PR this weekend.

[...]

PR #473, opened 2017-08-29
Contextless task

pekkaklarck:
Implements #445. User documentation missing but code hopefully OKish.


TaiSHiNet:
@pekkaklarck @bitprophet any chance this can be picked back up?
If there is coding help needed do say so and I'll do my best to aid

PR #482, opened 2017-10-12
Fixed bug where function equality compared text, not by ref...

mrdecav:
… so the same functions with the same names and same body (but different closures) would be equal (and thus deduped, amongst other bad things).

I've found that the previous implementation was creating issues when you have a complex build tree that may have the same boilerplate code in it (but using different global args)... See the test I added as an example of the failure.

I think the test for "useful equality" was broken (fixed here)... The reason equality didn't "work" is because they were in fact not equal functions—they were different references, and could close on different things.

### DIFF
--- a/invoke/tasks.py
+++ b/invoke/tasks.py
@@ -86,19 +86,8 @@ class Task(object):
     def __eq__(self, other):
         if self.name != other.name:
             return False
-        # Functions do not define __eq__ but func_code objects apparently do.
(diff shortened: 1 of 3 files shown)


bitprophet:
I acknowledge your use case, but worry there is another that caused us to implement the code being removed, which will now break (again.) We should look at git/ticket history to doublecheck this; I don't recall offhand exactly where or how many times we're caring about task equality.

PR #493, opened 2017-11-28
Fix #407 Windows shell compatibility bug

gucciferXCIV:
Fixes #407, based on @mkusz [solution](https://github.com/pyinvoke/invoke/issues/371#issuecomment-332458981) in #371, by changing `Config.global_defaults <invoke.config.Config.global_defaults>`. This is done first, by checking if the OS is Windows, and if so, second, changing the value of `defaults['run']['shell']` to the Windows default shell setting, the value of the environment variable `COMSPEC`.


gucciferXCIV:
Needs edit.

Issue #449, opened 2017-05-30
Positional arguments not described in help


cyounkins:
```
$ cat tasks.py
from invoke import task

@task
def hi(ctx, name):
    print("Hi {0}!".format(name))


$ invoke --help hi
Usage: inv[oke] [--core-opts] hi [--options] [other tasks here ...]

Docstring:
  none

Options:
  -n STRING, --name=STRING


$ invoke hi
'hi' did not receive all required positional arguments!
```

The auto-generated usage does not indicate that a positional argument is required. `hi [--options] [other tasks here ...]` seems to imply the opposite - that all arguments are optional.

In addition, we should be able to print the required positional arguments when we do not receive all of them.

bitprophet:
Good catch, this feels like a bug to me. Thanks!

dmulter:
+1

bitprophet:
To clarify, I think this change should both:

- add more info to `--help` output re: positional arguments
- display `--help` output in the situation causing us to complain about positional arguments, automatically (this comes up a lot actually.)

riverfr0zen:
I've posted a pull request for this issue at #504

gmmeyer:
I don't think the above commit fully fixes this bug. If you don't supply sufficient positional arguments, the help flag won't work properly, either. Here's an example:

If you have a task, something like this that was written by one of my coworkers, @ofek:

```python
@task(help={
    'package': 'The package to upgrade throughout the integrations',
    'version': 'The version of the package to pin',
    'verbose': 'Whether or not to produce output',
})
def upgrade(ctx, package, version, verbose=False):
    """
    Do Something
    """
    ...
```

And you do not supply the positional arguments, the task will just fail:
```
❯❯❯ invoke upgrade -h
'upgrade' did not receive all required positional arguments!
```

If you put help before the task, it will print out normally:
```
❯❯❯ invoke -h upgrade
Usage: inv[oke] [--core-opts] upgrade [--options] [other tasks here ...]

Docstring:
  Do Something

Options:
  -e, --verbose                 Whether or not to produce output
  -p STRING, --package=STRING   The package to upgrade throughout the integrations
  -v STRING, --version=STRING   The version of the package to pin
```

dongfengdejia:
May I ask is this one bug too? (i mean if there not have args, but there will raise error?)

code (invoke-0.22.1-py2.7.egg\invoke\tasks.py   function: def argspec(self, body)):
        ...
[...]

PR #504, opened 2018-02-07
solution for #449 

riverfr0zen:
These changes attempt to solve the issue by introducing a UserError subclass of ParseError that is thrown by parser.StateMachine instances when it is found that the invocation needs more positional args. Program instances look for UserError in the run() method and invoke their print_task_help(). 

The result for users is that the positional arg error is followed by the task help message.

Disclaimer: I just started using this package today, so my approach may be naive here.


anarcat:
i do not believe this PR fixes the issue it claims to, and recommend just closing it.

PR #505, opened 2018-02-10
Remove duplicate argument handling from task decorator

rpkilby:
As far as I can tell, it isn't necessary to perform argument validation in the `task` decorator, as the task is always instantiated immediately. 

- The positional args and `pre` kwarg check can be handled by `Task.__init__`, which will raise a `TypeError` for multiple values for `pre`. That said, this check can easily be added back. 
- Unknown kwargs should probably be handled by `Task.__init__`, which will raise a `TypeError`.

### DIFF
--- a/invoke/tasks.py
+++ b/invoke/tasks.py
@@ -26,8 +26,6 @@ class Task(object):
     """
     Core object representing an executable task & its argument specification.
     """
-    # TODO: store these kwarg defaults central, refer to those values both here
-    # and in @task.
     # TODO: allow central per-session / per-taskmodule control over some of
     # them, e.g. (auto_)positional, auto_shortflags.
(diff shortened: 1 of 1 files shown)


rpkilby:
For some reason the py36 build randomly failed..

bitprophet:
I just incidentally handled some of this (removal of the explicit TypeError check) this past week, and I put in a TODO for the rest of this (again, without remembering that this PR existed). tl;dr this is indeed dumb and and stupid and (partly) gets in the way of extending this class/decorator pair.

I'm worried there are some small, stupid mismatches in default values between what `@task` does and what `Task.__init__` does (e.g. a key in one defaulting to a tuple, and a list or None or whatever in the other). But that's not a great argument for leaving this around forever either. Gotta think it over a bit but gonna drop this in the first milestone so I remember to make a call.

PR #519, opened 2018-05-02
Add local_environ context manager.

steveprentice:
Also removed the unused environ test fixture which had similar functionality.

This is extracted from fabric/fabric#1738 and will be used in the Connection.socks_proxy() context manager.


steveprentice:
@bitprophet, do you have an ETA on when this might be merged? I'm waiting for this to merge before reopening the socks proxy PR in fabric.

PR #522, opened 2018-05-11
Make test fallback to system modules when vendorized one are not available

athmane:
This PR is needed for distro integration/packaging.
Thank you in advance.


bitprophet:
Can you rebase or recreate this against the 1.0 branch? Thanks!

PR #525, opened 2018-05-14
Fixup appveyor... maybe

rpkilby:
This temporarily disables Appveyor CI by moving the config into the `appveyor/` directory. 

I would attempt to fix this, but that would require making changes to Invocations. The problem is:
- invoke's tasks depend on invocations.
- invocations depends on blessings
- blessings is [not compatible with Windows](https://github.com/erikrose/blessings/issues/21), and therefore, Appveyor.


bitprophet:
Lazily not actually looking at appveyor yet...what's the status of this now?

If blessings straight up cannot be imported on Windows, that's one thing; if it's just that its runtime invocation blows up due to that lack of terminal size info (...wonder what we do? probably default to 80x24?) we could try skipping whatever we're doing with it on Windows conditionally.

thebjorn:
If all you need is screen size on windows, take a look at the ScreenInfo class here https://github.com/thebjorn/doscmd-screen/blob/master/screen.py  The docs for the rest of the module is at (https://doscmd-screen.readthedocs.io/). It's limited but x-platform..

bitprophet:
@thebjorn We ourselves handle screen size "well enough" on Windows currently (at least, it doesn't explode!), that's not the issue - the issue is what `blessings` (which invocations, our "extra stuff" package currently relies upon) does something internally which is not Windows compatible :)

We can't control that so we need to figure out how to guard against the problem one level up.

rpkilby:
As far as I can tell, blessings is just used for terminal colors, which could probably be replaced with `ansicolors` or a similar package.

https://github.com/pyinvoke/invocations/blob/e11e77ed63373f0d95e1813af3967e4a0101c6cb/invocations/packaging/release.py#L82-L94

bitprophet:
I'd have to doublecheck (I felt like there was at least one other thing blessings was used for besides ANSI) but yes, that's a possibility too - if blessings' kitchen sink approach is too problematic it might not be worth using long term.

Issue #552, opened 2018-07-08
Blocking when running commands looking for stdin close


plockc:
Example:
```
@task
def cat(c):
    c.run("cat")
```
```
echo Hi | timeout 3 invoke cat || echo failed
```
This prints Hi but never exits because cat never receives a close of stdin and is waiting for more, and invoke is waiting on the subprocess.

In a different issue, I notice that the console has trouble after running the above command 

Caused by two bugs:

First is that ExceptionHandlingThread.is_dead return expression is wrong, it should be "or" instead of "and":
https://github.com/pyinvoke/invoke/blob/master/invoke/util.py#L274

Second, is that the stdin for the process is not closed after receiving EOF:
https://github.com/pyinvoke/invoke/blob/master/invoke/runners.py#L659

Python 3.6.5
Ubuntu 18.04
Invoke 1.0.0.0

bitprophet:
Can confirm the lack of exiting easily enough.

Not getting terminal weirdness however, even under bash, which is where I am able to reproduce the potentially-related fabric/fabric#1812. Still trying to determine if the two issues stem from the same bugs.

bitprophet:
Still ongoing but that Fabric ticket seems to be caused by executing `with character_buffered(...)` multiple times concurrently (such that all but the first invocation erroneously snapshots, then restores, an already-character-buffered terminal state) which is what causes the weird local terminal behavior (specifically, the `echo` flag is set to off, and then never properly restored).

Given the code is a contextmanager, I don't see why it would ever hit that particular issue in single-threaded Invoke-only code; no matter what happens inside the block (early exit, error, process termination short of SIGKILL) the post-yield, end-of-block terminal restoration code (which is in a `finally` clause) should always run.

In addition, when (our) stdin is hooked up to a pipe, as it is here, that contextmanager should become a no-op. The detection of whether to do that has historically been a bit fragile, so maybe this exposes another hole in it. Either way, I can't reproduce that angle, but I can reproduce the hanging/never-exiting problem so I'll focus on that first and we can make a new issue for terminal state corruption if necessary.

plockc:
hi @bitprophet, the terminal weirdness only happened once, I can't reproduce, I probably should not have mentioned in this ticket.

I have locally ran 1.1.1 and still can reproduce the blocking issue, while taking my 1.0.0 patch and rebasing on top of 1.1.1 works for the "cat" test.

[...]

PR #553, opened 2018-07-08
Unblock subprocesses waiting for stdin close

plockc:
Partial fix for https://github.com/pyinvoke/invoke/issues/552

BREAKING CHANGE FOR RUNNER IMPLEMENTATIONS:
New method added close_proc_stdin to allow the parent Runner to ask
subclass to close the stream after it finds the end of the stream.

Fix ExceptionHandlingThread.is_dead which before would report dead when
thread was not alive and there was no exception info

This does not fix when pty=True, more significant changes are required for that case

Two test cases were adjusted that were wrongly asserting that the thread
should be "is_dead" after thread.join()


plockc:
The appveyor seems to fail on many (all?) jobs with failing to find:
C:\Python27\Scripts\spec.exe

invoke/runners.py
```diff
@@ -659,6 +659,8 @@ def handle_stdin(self, input_, output, echo):
                 elif data is not None:
                     # When reading from file-like objects that aren't "real"
                     # terminal streams, an empty byte signals EOF.
+                    if not self.using_pty:
+                        self._close_proc_stdin()
```
plockc: can check for the existence of attribute on self to allow subclasses to defer the fix

bitprophet:
I'm a bit leery about the `is_dead` change since you're changing the test to just match the new behavior. Been a while since I looked at this part of the code, but I'm _guessing_ the original author explicitly intended the `is_dead` flag to exhibit the properties it did in the test. Inverting the logic seems like a bad idea?

Can you provide more details about how the current test and implementation are both wrong? Thanks!

plockc:
Hi @bitprophet I have to run but had time to address the implementation side of things, can describe the test case issues later:

Original code:
```
return (not self.is_alive()) and self.exc_info is not None
```
New code:
```
alive = self.is_alive() and (self.exc_info is None)
        return not alive
```
For new code, if a is "is_alive" and b is "self.exc_info is None", we can substitute and then rearrange for the `is_dead` calculation:
```
!alive
!(a and b) 
!a or !b  # demorgan's
!(is_alive) or !(self.exc_info is None)
not self.is_alive or self.exc_info is not None
```

Which is exactly the same as the original except using `or` instead of `and`

-----

Consider the case where the thread completes normally,  the thread's is_alive will be False, in the _original_ code:
```
(not self.is_alive()) and self.exc_info is not None
(not False) and self.exc_info is not None
True and self.exc_info is not None
self.exc_info is not None
```
Which means: `is_dead == has exc_info` . . . but if we exit _normally_, there would be no exception captured, and so is_dead reports false, keeping the input thread alive

----

The test case was implemented incorrectly to make it pass, here is one of them:
https://github.com/plockc/invoke/blob/master/tests/concurrency.py#L39-L42
```
 t = EHThread(target=self.worker, args=[Queue()])
 t.start()
 t.join()
assert not t.is_dead
```
The last line is asserting that the thread is alive, right after a thread join, `thread.join()` docs: "Wait until the thread terminates"

plockc:
i updated the tests for additional validation of underlying conditions, and the call to the new close stdin is now checked first to allow subclasses in other repos (fabric?) to update on it's own schedule

plockc:
I didn't call it out here, but there is a minimal  test scenario in https://github.com/pyinvoke/invoke/issues/552

plockc:
To make things more interesting, occassionally (maybe 1/4 of the time), I get an UnexpectedExit error in program.py:run (running the same task) when I have `in_stream=StringIO(meta_data)`:
```
Encountered a bad command exit code!

Command: 'cd /var/lib/libvirt/images/base && sudo tee meta-data > /dev/null'

Exit code: None

Stdout: already printed

Stderr: already printed
```
I don't know yet if it's something I introduced, or pointing to something deeper and darker.

seansfkelley:
Tjhois

brianbolt:
Still reproducible in `fabric==2.7.1`

PR #557, opened 2018-07-27
Ensure dir change removed despite exceptions

code-tree:
Hi, so I don't believe the current implementation of `cd()` handles exceptions. Take for example:

```python
try:
    with c.cd('subdir'):
        c.run('ls')
        raise Exception()
finally:
    c.run('ls')
```

The final `run()` call is executed in the same dir as the first because the dir change wasn't popped.

Note: Please test before merging (I'm writing in github to save time)

### DIFF
--- a/invoke/context.py
+++ b/invoke/context.py
@@ -365,8 +365,10 @@ class Context(DataProxy):
         .. versionadded:: 1.0
         """
         self.command_cwds.append(path)
-        yield
-        self.command_cwds.pop()
+        try:
+            yield
+        finally:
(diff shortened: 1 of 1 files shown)


bitprophet:
Good catch, thanks! Throwing in bugfix milestone, I'll handle backporting/testing when I get to it.

bitprophet:
Looks like #640 extends this to handle the same issue in the sibling method, `cwd`, linking here for posterity. Merging these two now and adding tests/changelog/etc (😩)

PR #567, opened 2018-08-17
Rename link 'API Docs' to 'Documentation'

kenden:
The link API Docs is wrong, it's not just API documentation.
This should make the doc easier to find for new users


bitprophet:
I wanted to argue but honestly, you might be onto something. `www` is more "project metadata" and `docs` does include non-API documentation. Thanks! Let's try this.

Issue #461, opened 2017-07-28
Advanced task dependencies / caching


bitprophet:
## Use case description

Core desire is to strengthen relationships between tasks above/beyond what the existing pre/post functionality achieves; and to structure a nontrivial set of dependent state as a graph of tasks which only execute as many times as is required to set up that state.

Specific example is an existing nontrivial task tree around performing cloud automation, configuration management, and similar things. Many/most tasks in this tree are currently wrapped with a `@state` decorator whose body does a large amount of "heavy lifting":

- processes pre-existing configuration directives, including aborting if they aren't defined
  - A reminiscent predecessor is the old and (as far as I can tell) barely used [fabric.api.require](http://docs.fabfile.org/en/1.13/api/core/operations.html#fabric.operations.require)
- instantiate network and API clients using that early configuration (and/or pulling from a secrets storage API)
- generate new configuration settings based on runtime information (including data gathered from the above clients)
- attempts at each step to perform idempotency, including skipping expensive operations if necessary (so no re-connecting to already-connected APIs; no re-reading already-read files; etc.)

At the least, this decorator "wants" to be split up into a bunch of smaller parts, each of which are connected by what state they themselves depend on, with the outermost/topmost caller (i.e. the decorated CLI task) specifying more precisely which bits they need. (This already exists in part with some args passed into the decorator, e.g. `@state(limited=True)` skips everything but the first few config bits and an early API client.)

Those smaller parts would then be capable of memoizing their results, so that e.g. if they are declared to "generate a value for config path `c.aws.clients.vpc`", and that config path already exists/is non-empty when something starts calling that function, it simply records this fact and short-circuits.

Presumably one could expand this concept all the way to abstract things like "task A requires an admin cloud instance and a database" - including some sort of registry-driven "find me some task that satisfies needs XYZ, cuz I need X Y and Z" setup (though that feels much more fraught and complex, and is probably a distinct API feel from the "do exactly what i say but feel free to skip anything you already did" described above.)

## Solutions brainstorm

TK

----

Related, possibly subsumed tickets:

- #41 - the original dependencies ticket, long closed
- #45 - this is the closest match, and arguably a duplicate, though I interpret that ticket to be more "make the existing pre/post deduplicating use a DAG", whereas this ticket is about extending pre/post itself to have a richer API, even if that probably requires the DAG anyways.
[...]

PR #578, opened 2018-09-21
Added make-like checks to allow smarter deduping.

haydenflinner:
This implements checks as documented here - https://bit.ly/2MP86Ic - and on #461.
Note the current limitation that all checks must be tasks.

Here's an example usage:
```
from invoke import task
from os.path import exists

@task
def already_built(ctx):
    print("checking...")
    return exists('x')

@task(checks=[already_built])
def build(ctx):
    print("creating x")
    ctx.run('touch x')

@task(pre=[build])
def read_file(ctx):
    print("found x!")
```

What do you think of renaming `checks` to `skip_if`? And hope you don't mind the little bit of cleanup I did in tasks.py. Thanks for the awesome library!

### DIFF
--- a/invoke/executor.py
+++ b/invoke/executor.py
@@ -90,6 +90,7 @@ class Executor(object):
         # Obtain copy of directly-given tasks since they should sometimes
         # behave differently
         direct = list(calls)
+
         # Expand pre/post tasks
         # TODO: may make sense to bundle expansion & deduping now eh?
         expanded = self.expand_calls(calls)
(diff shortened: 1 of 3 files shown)


haydenflinner:
Some notes -- We could use this to remove all current de-duping efforts, and just provide an already_ran check ( or skip_if=[already_ran] if you will :wink:) that logs that the function has already been run either on the function or in `ctx`. I also made no formatting efforts :stuck_out_tongue:

haydenflinner:
@bitprophet Could you estimate how long it will take to merge this, if you're interested? I'd like to use it at work, but I don't really want to build on top of a patch that won't go anywhere. Check out the power that this feature gives you -- https://gist.github.com/haydenflinner/97eec5d3a0f84312bc221367acd3bfa2

haydenflinner:
I'll split it off this weekend if haven't heard anything. As an easy division, everything below here on GitHub (https://github.com/pyinvoke/invoke/pull/578/commits/06209adc91c966885ac143654d7f2122f0bb1908#diff-a2c23243533cad12946af2f1122bb4ebL49) is just the cleanup associated with the TODO that was there. Everything above is added for @checks.

haydenflinner:
@tuukkamustonen I didn't ever get around to splitting the PR, but I did release something called `magicinvoke`, check it out on PyPi :smile:

Issue #378, opened 2016-07-20
*args and **kwds support in tasks


pandey1010:
Is it possible to have support for args and/or kwds in tasks? I need it sometimes to pass all argument commands, as is, to another script.

bitprophet:
I feel like there's an open ticket for this already, but I can't find it. Semi-related issues that are not duplicates: #174, #9, #159.

----

Is it possible? Not at the moment, since it makes multi-task CLI parsing quite ambiguous; there's no obvious "one right way" to implement it. I.e.:

```
$ inv mytask a bunch of things I want to go into splat args here
```

You probably want that to act like:

``` python
mytask(*['a', 'bunch', 'of', 'etc'])
```

But what if your task collection has another task called `bunch`? It's then impossible to tell the above apart from an intent to call `mytask(*['a'])` followed by `bunch('of', 'things', 'etc')`.

Hopefully there's a way to square these two concerns; offhand:

* Just raise an exception if the user actually triggers the ambiguity; I *think* the parser can handle this, but it may not be able to, been a while since I was in there.
    * If this is possible, it's probably ideal, since the ambiguity case is likely to be pretty rare.
    * Might want to be a warning instead of an exception, since otherwise there's no way for a user to work around it - i.e. maybe they do have a `bunch` task, but they are not intending to ever call it on the same command line as `mytask` - they want the word `bunch` to be passed to `mytask` as a regular splat-arg member.
        * Or have a core flag like `--warn-about-splat-args`?
* Have the parser know beforehand that a task takes arbitrary args (either by an explicit task-generation parameter, or just by noticing the existence of splat-args in the signature), and have it explicitly **disable** any subsequent multi-task lookup.
    * This is only good if we assume users always intend to be using the splat-args and aren't actually making a silly mistake.
    * It also violates [Zen of Python #12](http://zen-of-python.info/in-the-face-of-ambiguity-refuse-the-temptation-to-guess.html#12) so I'd probably rather not. Maybe let users enable this with a setting/flag.

SwampFalc:
Since I could usefully use an *args for a project I'm working on, I'd like to add my vote for this feature.

If this feature were to come at the cost of being able to have multiple tasks in one invocation, then I personally wouldn't mind. There are many ways in which you can chain tasks, but there is no other way to have an arbitrary amount of arguments for a task.

Otherwise, one other option is that, if the parser knows about splat-args, then it will accept multiple instances of the same argument.

[...]

PR #582, opened 2018-09-23
Stop parsecontexts deepcopy. Add *args **kwargs support.

haydenflinner:
First step towards #378. See (#TODO 378)s.

### DIFF
--- a/invoke/collection.py
+++ b/invoke/collection.py
@@ -404,9 +404,10 @@ class Collection(object):
         result = []
         for primary, aliases in six.iteritems(self.task_names):
             task = self[primary]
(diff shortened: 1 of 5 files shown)


therrj:
Stumbled across https://github.com/pyinvoke/invoke/issues/378 which led me here and I managed to get this merged and working with the latest invoke release (1.6.0).

There appears to be a bug with all of the "eat_all" calls as that appears to be a function but the code just does self.context.eat_all which would return the bound method signature and doesn't actually run the function. I don't know if this worked in the past, but I had to make those actual function calls ("eat_all()") and that resolved the issue and appears to work fine with 1.6.0

Issue #454, opened 2017-07-05
Accept pathlib.Path object in Context.cd


jimm-domingo:
Allow the path parameter to be a `pathlib.Path` instance as well as a string parameter.  As a convenience as `pathlib.Path` are used more and more for path handling.  Thanks.

bitprophet:
On one hand, I'm not sure how we could go about this without actually depending on `pathlib` existing. On the other, I do use `pathlib` or similar libs more often lately myself, and wouldn't be super opposed to using it internally where sensible, which would then open us up to vendoring it & thus enabling this feature.

brutus:
I just stumbled upon this while investigating a somewhat misleading error message:

```
$ inv isort
checking import statement order in '/home/brutus/Projekte/LinterTest/src'…
Traceback (most recent call last):
  File "/home/brutus/Projekte/LinterTest/.venv/lib/python3.6/site-packages/invoke/config.py", line 97, in __getattr__
    return self._get(key)
  File "/home/brutus/Projekte/LinterTest/.venv/lib/python3.6/site-packages/invoke/config.py", line 164, in _get
    value = self._config[key]
  File "/home/brutus/Projekte/LinterTest/.venv/lib/python3.6/site-packages/invoke/config.py", line 153, in __getitem__
    return self._get(key)
  File "/home/brutus/Projekte/LinterTest/.venv/lib/python3.6/site-packages/invoke/config.py", line 164, in _get
    value = self._config[key]
KeyError: 'cwd'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/brutus/Projekte/LinterTest/.venv/bin/inv", line 11, in <module>
    load_entry_point('invoke==0.20.2', 'console_scripts', 'inv')()
  File "/home/brutus/Projekte/LinterTest/.venv/lib/python3.6/site-packages/invoke/program.py", line 294, in run
    self.execute()
  File "/home/brutus/Projekte/LinterTest/.venv/lib/python3.6/site-packages/invoke/program.py", line 409, in execute
    executor.execute(*self.tasks)
  File "/home/brutus/Projekte/LinterTest/.venv/lib/python3.6/site-packages/invoke/executor.py", line 129, in execute
    result = call.task(*args, **call.kwargs)
  File "/home/brutus/Projekte/LinterTest/.venv/lib/python3.6/site-packages/invoke/tasks.py", line 109, in __call__
    result = self.body(*args, **kwargs)
[...]

PR #583, opened 2018-09-25
allow Pathlib.path objects to be used with Context.cd()

fndari:
This addresses #454, implementing the simplest cast-to-`str` solution as originally suggested by @brutus, plus a basic unit test.


bitprophet:
Rolling into #607

Issue #592, opened 2018-10-25
pylint throws a DeprecationWarning for invoke


marcus-crane:
I was recently porting a work project to Python 3.7 and decided to upgrade our dependencies along the way. We use `pytest`, as does invoke, and it turns out v3.8 now bubbles up deprecation warnings.

We use invoke as a test runner, as does invoke itself, and this means we started getting a stream of `DeprecationWarning` messages due to one use of `inspect.getargspec` which has apparently been deprecated since Python 3.0.

`inspect.getfullargspec` is apparently a drop in replacement so I've done a PR that updates it. The test suite passes so I imagine it should just work fine with the addition of not outputting warnings to any dependents who use pytest 3.8 and above.

Let me know if there's anything else I'm missing.

Thanks!

bitprophet:
2.0.0 will be out soon & is now using `inspect.signature`!

PR #593, opened 2018-10-25
Fix DeprecationWarning output by newer versions of pytest

marcus-crane:
You can see an explaination of the issue that this PR attempts to solve here: #592 

Python 3.0 and above now uses the `inspect.getfullargspec` method as `inspect.getargspec` is deprecated in 3+.

Normally it wouldn't be an issue but new versions of Pytest bubble up DeprecationWarnings and as such, I noticed a stream of DeprecationWarnings inherited from invoke in one of our work projects.

### DIFF
--- a/invoke/tasks.py
+++ b/invoke/tasks.py
@@ -151,7 +151,7 @@ class Task(object):
         # TODO: __call__ exhibits the 'self' arg; do we manually nix 1st result
         # in argspec, or is there a way to get the "really callable" spec?
         func = body if isinstance(body, types.FunctionType) else body.__call__
-        spec = inspect.getargspec(func)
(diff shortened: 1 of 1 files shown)


marcus-crane:
Ah, silly me. I ran the tests using Python 3 and they worked but of course, that methods didn't exist for 2.7 which I forgot invoke also supports. I'll see if I can update it so it works seamlessly on both platforms.

marcus-crane:
@bitprophet I notice that Appveyor is hanging?

johnraz:
@marcus-crane could you rebase this on latest master? I wonder if that would fix the CI as more recent PR are passing. Thanks!

PR #600, opened 2018-11-13
copy in-memory in_stream like StringIO efficiently in chunks

sprohaska:
Previously, in_stream was processed one byte at a time even for
in-memory file-like objects such as io.StringIO, which can be safely
read in chunks without blocking.  In-memory file-like objects are now
efficiently copied in chunks.

Signed-off-by: Steffen Prohaska <prohaska@zib.de>

### DIFF
--- a/invoke/terminals.py
+++ b/invoke/terminals.py
@@ -214,9 +214,8 @@ def bytes_to_read(input_):
     Query stream ``input_`` to see how many bytes may be readable.
 
     .. note::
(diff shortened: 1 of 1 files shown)


NotSqrt:
While this PR is not merged, it's possible to do:
```python
from unittest.mock import patch
with patch('invoke.runners.bytes_to_read', return_value=8196):
    # here your code that uses in_stream & StringIO/BytesIO
```

PR #604, opened 2018-12-27
Exclude all __pycache__/pyc/pyo files from sdist (fix #586)

### LINKED ISSUE #586: __pycache__ directory included in source tarball
The 1.2.0 release up on PyPI contains python bytecode in the source tarball:

 ```
$ mkdir /tmp/invoke-dl/ && cd /tmp/invoke-dl
$ wget https://files.pythonhosted.org/packages/ef/80/cef14194e2dd62582cc0a4f5f2db78fb00de3ba5d1bc0e50897b398ea984/invoke-1.2.0.tar.gz
$ tar xzf invoke-1.2.0.tar.gz
$ find invoke-1.2.0 -name \*.pyc
invoke-1.2.0/invoke/completion/__pycache__/complete.cpython-36.pyc
invoke-1.2.0/invoke/completion/__pycache__/__init__.cpython-36.pyc
```

These are, transitively, included in the wheel files too.

Please make sure to not include `__pycache__` directories in the source tarball.  I suspect that it ends up in the source tarball due to the following line in `MANIFEST.in`: https://github.com/pyinvoke/invoke/blob/e4bbd9b0d03cb3dbbbe2d8094f6d84a31fe51ed5/MANIFEST.in#L4

😞 

I suspect that all it takes is to add

```
recursive-exclude * __pycache__
recursive-exclude * *.py[co]
```
to `MANIFEST.in` to prevent this.

florisla:
Fixes #586 .

The best I could find on `recursive-exclude` documentation was https://docs.python.org/2/distutils/sourcedist.html#commands, and it wasn't really helpful.

This syntax works; I've verified it with bogus `*.pyc` and `*.pyo` files and `__pycache__` folder.  It does not strip more than it has to from the `.tar.gz`.

Tested on Debian 9.5 / Python 3.6 and Windows 7 / Python 3.7.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -7,4 +7,6 @@ recursive-exclude sites/*/_build *
 include dev-requirements.txt
 include tasks-requirements.txt
 recursive-include tests *
-recursive-exclude tests *.pyc *.pyo
+recursive-exclude * *.pyc *.pyo
+recursive-exclude **/__pycache__ *
+


florisla:
Rebased on master.

bitprophet:
Cherry picked back to 1.3.x. Thanks!

Issue #584, opened 2018-09-26
The FAQ document should mention renaming fabfile


dannystaple:
When coming to upgrade (sidegrade) the FAQ should declare, pretty early that the old "fabfile.py" should now be "tasks.py". 

This error message isn't clear:

`Can't find any collection named 'tasks'!`

Perhaps in `faq.rst.txt`:
```
Defining/executing tasks
========================

.. _fabfile:

`Can't find any collection named 'tasks'!`
--------------------------------------------

Tasks should be placed in a file named `tasks.py`. If you are converting from Fabric 1.x, you will
need to rename this as you make the other upgrade changes.
```

PR #608, opened 2019-01-06
FAQ: Tell Fab v1 users to rename fabfile to tasks.py

florisla:
This is a pull-request version of #584.


florisla:
Rebased on master.

PR #580, opened 2018-09-22
Soften some edges on help text behavior #409, #398

haydenflinner:
Fixes #409

PR #607, opened 2019-01-06
Make Context.cd() accept Path objects

florisla:
This is a rebase (with minor modifications) of #583.

The original 583 stalled on AppVeyor, but that is disabled now so this should pass CI.

Closes #583 - original PR
Closes #681 - more recent but similar PR
Fixes #454 - original issue
Fixes #577 - duplicate issue
Fixes #658 - duplicate issue

PR #611, opened 2019-01-07
Soften some edges on help text behavior (rebooted)

### LINKED ISSUE #398: Mismatched help keys and task args doesn't result in a useful error
Hi,
I am trying to add help text to a task as follow: (Pyinvoke 0.13.0)

```
@task(help={'name': "Use --type aj and --type mp"})
      def build(ctx, type=None):
         pass
```

The issue is invoke --help build prints nothing in `OPTIONS`.
The check `if name in self.help:` (Line: 186) inside function `def arg_opts()` in invoke/tasks.py is getting failed and not value gets assigned to `opts['help']` that's what i noticed. Any workaround for this issue?

### LINKED ISSUE #409: help text doesn't work properly in parameters with underscore
The next code doesn't generate a help text:
```
@task(help={'include_author': 'Include author in the footer'})
def docs(ctx, include_author=False):
    """
    Docs generator
    """
    print("Documenting...")
```

```
$ inv --help docs
Usage: inv[oke] [--core-opts] docs [--options] [other tasks here ...]

Docstring:
  Docs generator

Options:
  -i, --include-author
```

But replacing the key "include_author" with "include-author" in the help dictionary works:
```
$ inv --help docs
Usage: inv[oke] [--core-opts] docs [--options] [other tasks here ...]

Docstring:
  Docs generator

Options:
  -i, --include-author   Include author in the footer
```

### LINKED ISSUE #533: Help does not display if the argument contains an underscore
Pretty much the title. Here is how to reproduce it:

```
@task(help={
    'service': "The service that should be deployed",
    'env': 'Target environment',
    'git_branch': '[Optional] Name of the branch to be used in the host machines',
})
def deploy(c, service, env, git_branch=None):
    """
    Deploy a service into a specific environment and optionally using a specific git branch
    """
```
The result of `invoke -h deploy`:

```
Options:
  -e STRING, --env=STRING          Target environment
  -g STRING, --git-branch=STRING
  -s STRING, --service=STRING      The service that should be deployed
```

Also, I have no idea why the options are not listed in the right order!

### LINKED ISSUE #590: @task(help={...}) should be dash/underscore agnostic
This doesn't work:

```python
@task(help={'foo_bar': '...'})
def some_task(ctx, foo_bar):
    ...
```

This works:

```python
@task(help={'foo-bar': '...'})
def some_task(ctx, foo_bar):
    ...
```

It's kinda confusing that you need to use the dashed syntax in `help`, while the parameter is really named with underscore.

florisla:
This is a rebase of #580, with added tests.

The original #580 failed on Travis due to code formatting (`black`).

New tests:

    Program
        help
            per parameter
                displays when help is supplied with dashes in key
                displays when help is supplied with underscores in key

    Collection
        to_contexts
            raises ValueError on help without matching parameter

Fixes #398 Bug report: no error on mismatched help keys and task arguments
Fixes #409 Bug report: help text on parameters with underscore
Closes #533 Duplicate bug report to #409
Closes #590 Duplicate feature request to #409
Closes #580 Original merge request

### DIFF
--- a/invoke/tasks.py
+++ b/invoke/tasks.py
@@ -256,7 +256,9 @@ class Task(object):
         if self.help:
             raise ValueError(
                 "Help field was set for params that didn't exist: {}".format(
-                    list(self.help.keys())))
+                    list(self.help.keys())
+                )
+            )
         return args
 
 
--- a/invoke/tasks.py
+++ b/invoke/tasks.py
@@ -217,7 +217,7 @@ class Task(object):
         # Help
         help_name_key = name if name in self.help else opts["attr_name"]
         if help_name_key in self.help:
-            opts["help"] = self.help[help_name_key]
+            opts["help"] = self.help.pop(help_name_key)
         return opts
 
     def get_arguments(self):
@@ -252,6 +252,11 @@ class Task(object):
                 if arg.name == posarg:
                     args.insert(0, args.pop(i))
                     break
+
+        if self.help:
+            raise ValueError(
+                "Help field was set for params that didn't exist: {}".format(
+                    list(self.help.keys())))
         return args
 
 
--- a/invoke/tasks.py
+++ b/invoke/tasks.py
@@ -215,7 +215,9 @@ class Task(object):
                 opts["kind"] = kind
             opts["default"] = default
         # Help
-        help_name_key = name if name in self.help else opts["attr_name"]
+        help_name_key = (name if name in self.help
+                         else opts["attr_name"] if "attr_name" in opts
+                         else None)
         if help_name_key in self.help:
             opts["help"] = self.help.pop(help_name_key)
         return opts
--- a/invoke/tasks.py
+++ b/invoke/tasks.py
@@ -215,8 +215,9 @@ class Task(object):
                 opts["kind"] = kind
             opts["default"] = default
         # Help
-        if name in self.help:
-            opts["help"] = self.help[name]
+        help_name_key = name if name in self.help else opts["attr_name"]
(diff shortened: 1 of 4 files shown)


florisla:
Sorry for opening so many pull requests recently, it's not my goal to overload the maintainer(s) with work.

My main motivation was fixing #409 and the others are just attempts to get some low-hanging-fruit issues closed.

florisla:
Rebased on master since I spotted #409 was in the 'Next bugfix' milestone.

Plug: #607 would be a nice one to include as well.

invoke/tasks.py
```diff
@@ -215,8 +215,9 @@ def arg_opts(self, name, default, taken_names):
                 opts["kind"] = kind
             opts["default"] = default
         # Help
-        if name in self.help:
-            opts["help"] = self.help[name]
+        help_name_key = name if name in self.help else opts.get("attr_name")
+        if help_name_key in self.help:
+            opts["help"] = self.help.pop(help_name_key)
```
haydenflinner, line 220: On line 61 above, we could perhaps use `self.help = help.copy() if help else {}` rather than help directly, in case someone is passing the same help dictionary to multiple decorators to avoid copy-pasting

bitprophet: About to read all this over myself, but I think I agree; that sort of defensive/functional-style programming is something I am frequently worrying about but not always finding the time to actually implement 🙃

bitprophet:
Looking at this now with an eye for getting it out in 1.5.

PR #613, opened 2019-01-16
Allow tasks to invoke other tasks applying task decorators, similar to fabric1 `execute`

### LINKED ISSUE #170: Easy execution of tasks from within Python itself

rectalogic:
Fixes #170

This seems to satisfy my use case of execute in fabric1.

Example fabric2 fabfile.py
```python
from fabric import task
from fabric.main import program


@task(hosts=["ubuntu@playapp1", "ubuntu@playapp2"])
def uname2(c):
    print("run uname2", c.host)
    c.run("uname -a")


@task(hosts=["ubuntu@playapp2"])
def uname(c):
    print("run uname", c.host)
    c.run("uname -a")


@task
def executeme(c):
    print("run executeme task")
    program.invoke_execute(c, "uname2")
    program.invoke_execute(c, "uname")
```


haydenflinner:
I went ahead and merged this in my extension, `magicinvoke`. You should be able to `pip install magicinvoke` over top of your current `invoke` install and pretend this has been merged upstream. But note that I renamed it from `invoke_execute` to just `invoke` :smiley:. You might also like [this decorator](https://magicinvoke.readthedocs.io/en/latest/api/magicinvoke.html#magicinvoke.get_params_from_ctx) which will allow you to enforce that parameters are in ctx by making them required parameters in the function header, rather than waiting for the eventual _ctx has no such key_ error.

breisig:
Has this been merged yet?

PR #629, opened 2019-03-30
Fallback to system lib if vendorized one does not exists

athmane:
Ref to #522


athmane:
closing since GH is showing 'unknown repository'

Issue #554, opened 2018-07-09
default value for positional argument


antoine-gallix:
I've been playing with `optional` and `positional` arguments of the `task` decorator, but I cannot find a way to achieve to following. I have a function that run scripts that are placed in a folder. I want to be able to run scripts by name, and list the available scripts. I'd like to achieve the following:

```
inv run_script script_name
> run the given script
inv run_script --list
> print list of available scripts
```

Up to now, I could only manage this with an optional argument:

```
inv run_script --name script_name
> run the given script
inv run_script --list
> print list of available scripts
```

My question is : How can I make an argument be both positional, and have a default value?

plockc:
from the docs (http://www.pyinvoke.org/), you can run multiple tasks with one `invoke` command, which I think means no positional arguments:
"From GNU Make, it inherits an emphasis on minimal boilerplate for common patterns and the ability to run multiple tasks in a single invocation:"
`$ invoke clean build`

bitprophet:
@plockc Not quite, there is both implicit and explicit support for positional arguments - but the multi-task nature of the parser is why it has to be handled carefully so it knows when a token is a positional argument value, and when it should signal looking for the next task 'context'.

@antoine-gallix I _think_ what you're describing -- a positional but also optional (because it has a default value) argument -- falls under an outstanding issue, #159, which is somewhere in the middle of my "top issues" priority list. Going to close this as a duplicate on those grounds but feel free to correct me if I'm not looking closely enough. Thanks!

plockc:
I hadn't played with the positional arguments to Task yet and didn't read clearly enough, my bad @antoine-gallix

antoine-gallix:
Just for info, the collection default task option finally also fits my requirements:

```
inv run.script script_name
> run the given script. default task for "script" collection
inv run.script.list
> print list of available scripts
```

jlugao:
hey @antoine-gallix I am looking for exactly this, could you share a gist of how you've set up the project? I am new to invoke and couldn't figure out for myself
I was using a makefile to manage my projects shortcuts, but the argument passing was not smooth so I am migrating to Invoke

antoine-gallix:
@jlugao 
```python
namespace = Collection()

script = Collection('script')


@task
def list_scripts(ctx):
	....

script.add_task(list_db_scripts, 'list')

@task
def run_script(ctx, name):
	....

script.add_task(run_script, 'run', default=True)

namespace.add_collection(script)
```

```shell
[...]

PR #635, opened 2019-04-16
Support for optional positional arguments

### LINKED ISSUE #159: optional positional argument
If I understand correctly (based on http://invoke.readthedocs.org/en/latest/api/tasks.html#invoke.tasks.task) optional AND positional arguments are not available (at the same time).

Any combination I tried I didn't manage to get both of these lines working:

``` bash
invoke build
invoke build /path/target.tar.gz
```

``` python
@target
def build(target='default.tar.gz'):
    print target
```

I think only optional arguments that are not positional work:

``` bash
invoke build --target=/path/target.tar.gz
```

Is there a way to achieve this? Or can you consider this feature request?

Thanks

olance:
This is an attempt at providing support for "optional posargs", i.e. positional arguments that can have a default value so that they can conveniently be omitted from the command line.

Fixes #159 and #554 (and possible other duplicate issues)

I've added some integration tests, but unit-testing the changes made in `ParserContext` and `ParseMachine` might be good too. (Note though that all existing tests pass without any change)

The implementation supports two ways of declaring an optional posarg:

**1. Optional arg explicitly declared positional**

```python
@task(positional=["name"])
def print_hello(c, name="World"):
    print("Hello, {}!".format(name))
```

This task can be called like so:

```
inv print-hello
> Hello, World!

inv print-hello Olivier
> Hello, Olivier!

inv print-hello --name Olivier
> Hello, Olivier!
```

**2. Positional arg explicitly declared optional**

Useful when there are several positional args and you don't want to list them all.  
Of course if you also have several optional args, then somehow you'll have to write down a full list of args 🙂 

```python
@task(optional=["second"])
def print_addition(c, first, second):
    if not second:
        second = 2

    print(int(first) + int(second))
```

This task can be called like so:

```
inv print-addition 3
> 5

inv print-addition 3 3
> 6

inv print-addition 3 --second 5
> 8
```

### DIFF
--- a/integration/main.py
+++ b/integration/main.py
@@ -62,6 +62,30 @@ class Main:
             _output_eq("inv print-name --name whatevs", "whatevs\n")
 
         @trap
+        def invocation_with_optional_positional_arg_as_positional(self):
+            _output_eq("inv -d print-hello Olivier", "Hello, Olivier!\n")
+
+        @trap
+        def invocation_with_optional_positional_arg_as_flag(self):
+            _output_eq("inv print-hello --name Olivier", "Hello, Olivier!\n")
+
+        @trap
+        def invocation_with_optional_positional_arg_no_value(self):
+            _output_eq("inv print-hello", "Hello, World!\n")
+
+        @trap
+        def invocation_with_multiple_positional_arg_single_optional(self):
+            _output_eq("inv print-addition 3 3", "6\n")
+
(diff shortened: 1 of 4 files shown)


olance:
Oh and I should not that this makes invoking multiple tasks at once impossible, which might be a show stopper? 😕

PR #640, opened 2019-05-13
Fix Context.cd and .prefix behaviour on exception

cytadela8:
When exception happens inside with statement it's escalated to the
contextmanager from with statement. By placing `self.command_cwds.pop()'
in finally block we make sure command prefixes are handled correctly on
exceptions in user code.

### DIFF
--- a/invoke/context.py
+++ b/invoke/context.py
@@ -296,8 +296,10 @@ class Context(DataProxy):
         .. versionadded:: 1.0
         """
         self.command_prefixes.append(command)
-        yield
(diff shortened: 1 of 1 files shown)


omarmarquez:
+1

bitprophet:
Thanks for this, #557 beat you to the `with cd` fix but missed the sibling change to `with prefix` 👍 ✨

omarmarquez:
Great that this have been fixed. When can expect it to be merged and released?

PR #653, opened 2019-08-01
Fixed i18n error on config load exception

### LINKED ISSUE #384: Config loading and locale bug

cybiere:
Fixes #384 using errno numeric value instead of strerror localized value

### DIFF
--- a/invoke/config.py
+++ b/invoke/config.py
@@ -889,8 +889,7 @@ class Config(DataProxy):
                 break
             # Typically means 'no such file', so just note & skip past.
             except IOError as e:
-                # TODO: is there a better / x-platform way to detect this?
-                if "No such file" in e.strerror:
+                if e.errno == 2:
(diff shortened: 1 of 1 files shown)


bitprophet:
Cherry-picked, changelogged, tested. thanks!

Issue #660, opened 2019-09-05
ctx.run fails due to bad command exit code


flazzarini:
### Scenario

I stumbled upon this issue when using [fabric2](https://www.fabfile.org) where I had a task which would execute [alembic](https://alembic.sqlalchemy.org/en/latest/) to upgrade/downgrade multiple local databases. When executing the task it would sometimes (completely randomly) fail with the following exception.

Please note the following is only happening when ``pty`` is set to ``False``.

Version of invoke used 1.3.0.

<pre>
Encountered a bad command exit code!

Command: './env/bin/alembic upgrade head'

Exit code: None

Stdout: already printed

Stderr: already printed
</pre>

### Reproduction of the bug

I replicated the fabric task to a pure invoke task like shown here. When executing the following code I get randomly the error mentioned above. I will execute the task multiple times to see that the error occurs in different runs every time.

#### Code
```
from invoke import task


@task
def upgrade(ctx):
    for index in range(1, 5):
        try:
            print("Run upgrade")
            ctx.run("./env/bin/alembic upgrade head")
        except Exception as exc:
            print(exc)
```

#### First run

```
$ ./env/bin/invoke upgrade
Run upgrade 1
Run upgrade 2
Run upgrade 3
Run upgrade 4
Run upgrade 5
Encountered a bad command exit code!

Command: './env/bin/alembic upgrade head'

Exit code: None

Stdout: already printed

Stderr: already printed


Run upgrade 6
Encountered a bad command exit code!

Command: './env/bin/alembic upgrade head'

Exit code: None

Stdout: already printed

Stderr: already printed


Run upgrade 7
Encountered a bad command exit code!

Command: './env/bin/alembic upgrade head'

Exit code: None

Stdout: already printed

Stderr: already printed


Run upgrade 8
Run upgrade 9
```

#### Second run

```
$ ./env/bin/invoke upgrade
Run upgrade 1
Encountered a bad command exit code!

Command: './env/bin/alembic upgrade head'

Exit code: None

Stdout: already printed

Stderr: already printed


Run upgrade 2
Run upgrade 3
Encountered a bad command exit code!

Command: './env/bin/alembic upgrade head'

Exit code: None

Stdout: already printed

Stderr: already printed


Run upgrade 4
Run upgrade 5
Encountered a bad command exit code!

Command: './env/bin/alembic upgrade head'

Exit code: None

Stdout: already printed

Stderr: already printed


Run upgrade 6
Encountered a bad command exit code!

[...]

PR #661, opened 2019-09-05
Give the process a chance to terminate correctly

flazzarini:
Should fix the issue described in #660

### DIFF
--- a/invoke/runners.py
+++ b/invoke/runners.py
@@ -1188,6 +1188,7 @@ class Local(Runner):
             return code
             # TODO: do we care about WIFSTOPPED? Maybe someday?
         else:
+            self.process.wait(timeout=3)
             return self.process.returncode
 
     def stop(self):


florisla:
The CI build fails because `Popen.wait()` does not have a `timeout` argument in Python 2.7.

See https://docs.python.org/2/library/subprocess.html#subprocess.Popen.wait

Would it be dangerous to simply call `wait()` in case of python <3 ?

flazzarini:
Yes I didn't think of python 2 support at all, but as python 2 is end of life on 01.01.2020 will you also drop support for it in the invoke library? If this is the case this fix could applied on a python3 only branch.

ralphie0112358:
I'd love to have this merged and released!  Any updates???

gbansaghi:
I would argue that `wait()` with no timeout would be the correct approach. In practice, 3 seconds should be enough, but the library shouldn't rely on such an arbitrary timeout value (what if it's run on a _very_ old machine, what if there's really heavy I/O, etc.)

bitprophet:
Thanks for this & the additional reports!

I can't drop Python 2 just yet, and I'm also just uncomfortable with a timeout-driven solution, I'd rather attempt to identify the race condition in play & solve it with a threading event or similar; so I'm probably going to supply my own patch instead of merging this directly (tho I reserve the right to backpedal furiously if that goes nowhere 😂 ). Please follow #660 for details.

Issue #101, opened 2013-12-17
Configure `run` in Collection


saulshanabrook:
I would love to be able to configure `Runner` options, like `echo=true`, on a `Collection`. 

Is this possible?

```
from invoke import Collection

from . import setup, reset


ns = Collection(setup, reset)

ns.configure({'config':'value'})
# would like to do:
ns.configure({'config':'value'}, run={'echo':True})


```

PR #663, opened 2019-09-09
Color output

exhuma:
This PR adds a very simple implementation for ANSI color output.

I am aware about pyinvoke/invoke#101 and that there are plans to "vendorise" something like `colorama`. This PR serves mainly to get the ball rolling and to scratch my own itch.

I am willing to incorporate `colorama` instead of using hand-crafted code.

However, I am unaware what is meant by "vendorising" code. Does that mean adding it as a simple dependency? Or rather copying the code *into* `invoke`. In the latter case, should I simply copy the code into `invoke/vendor`?

Let me know what option you'd prefer and I'll get to it.

### DIFF
--- a/invoke/runners.py
+++ b/invoke/runners.py
@@ -10,6 +10,7 @@ import time
 import signal
 
 from .util import six
+from .termcolor import white
 
 # Import some platform-specific things at top level so they can be mocked for
 # tests.
@@ -301,7 +302,7 @@ class Runner(object):
         env = self.generate_env(opts["env"], opts["replace_env"])
         # Echo running command
         if opts["echo"]:
-            print("\033[1;37m{}\033[0m".format(command))
+            print(white(command, bold=True))
         # If dry-run, stop here.
         if opts["dry"]:
             return self.generate_result(
--- a/invoke/termcolor.py
+++ b/invoke/termcolor.py
@@ -14,13 +14,17 @@ Example usage::
     >>> print(green('Hello World!', bold=True))
 """
 from os import getenv, isatty
+from random import choice
 from sys import platform, stdout
-from typing import Callable
+from typing import Callable, List
 from typing.io import IO
-
 
 #: If this is set to "True", no color output will be enabled
 DISABLE_COLORS = bool(getenv('INVOKE_DISABLE_COLORS', False))
+
+#: This collection keeps track of already assigned colorisers, so the same can
+#: be used repeatedly for the same key
+KEYED_COLORISERS = {}  # type: Dict[str, Callable[..., str]
 
 
 def color_wrapper(color_code: int) -> Callable[..., str]:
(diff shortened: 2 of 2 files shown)


exhuma:
Note that this also tries to reproduce the behaviour from fabric 1 (same function names, and same args as the `clr` module). However, the environmen variable `FABRIC_DISABLE_COLORS` has been renamed to `INVOKE_DISABLE_COLORS`.

One could argue to add a code into `fabric` along these lines:

    import invoke.termcolor
    import os
    invoke.termcolor.INVOKE_DISABLE_COLORS = os.getenv(
        'FABRIC_DISABLE_COLORS', False)

... although that looks a bit hacky, but would reproduce fabric-1 behaviour completely. Thoughts?

exhuma:
Sorry for the commit-spam... I am having issues with my dev environment and rely on travis.

exhuma:
There seems to be something strange going on with `sys.stderr` and `sys.stdout`. Maybe related to the `@trap` decorator from `pytest-relaxed`?

For the time being I will split the test `Runner_.command_echoing.uses_ansi_bold` as it's behaviour is really unpredictable. It works different in Python 2 and 3 as it seems and I don't really know why. Also, on my environment it behaves differently than on travis. Splitting it for Python 2/3 should make the tests pass... I hope 🤞

exhuma:
Okay... should be good now... all checks are passing at least.

I cleaned up the intermediary commits. Most of which are caused because I was unable to run the tests locally so I was dependent on pushing to make it run in Travis.

exhuma:
oh no... 😆

exhuma:
Done.

nickatnight:
When can we merge this in?

ofek:
@bitprophet Hello again! Are there still plans to support colored output?

bitprophet:
Thanks for this! Haven't dig into it yet but to answer the question about vendoring, it means adding something to the vendor folder like the rest of the stuff in there. There's actually an invoke task somewhere for pulling a pip installable library into that tree, tho I haven't used it in ages.

When I have time to review I'll make the judgement call re: whether it is worth replacing custom color codes with a lib or not.

exhuma:
Sure. No worries. Take your time. I'm not in a hurry. If you need any feedback, don't hesitate to give me a nudge.

I also see that there are conflicts on the branch. If I don't forget about this I'll try to remove those to make merging this a bit less painful :wink:

neozenith:
Since the inception of this PR there is another library that is heavily supporting colorising terminal text across the multitudes of terminals:
https://github.com/Textualize/rich

Even `pip` itself vendored it:
https://github.com/pypa/pip/issues/10423

This might be a good feature to target in the v2.0 of invoke once Python2 support is dropped and all the work on CircleCI is tidied up as well as some outstanding bugs for python 3.10

The colorised output could be a good visual indicator of the major change to invoke that it aesthetically _got a coat of paint_.

Issue #555, opened 2018-07-13
simple decorator doesn't seem to work


lsh-0:
The `@task` decorator doesn't appear to play nicely with other decorators, unlike Fabric. I may not have read this part of the documentation yet, but I wouldn't expect @task to undermine python norms.

For example:
```python
def capture_error(fn):
    @wraps(fn)
    def wrapper(*args, **kwargs):
        try:
            return fn(*args, **kwargs)
        except AssertionError as err:
            print(err)
    return wrapper

@task
@capture_error
def foo(c):
    print('bar')
    raise AssertionError('baz')
```

this looks fine:

```
inv foo
>>> bar
>>> baz
```

but when I add a parameter:

```python
@task
@capture_error
def foo(c, p1):
    print(p1)
    raise AssertionError('baz')
```

I get:

```
inv foo bar
>>> No idea what 'bar' is!
```

The parameter handling docs offer no insight into what the problem is.

bitprophet:
This is likely an oversight / lack of porting those extra careful parts of Fabric 1's decorator. #246 seems related and may fix it (though it's now kinda old so may need a revisit). Gonna leave this open and in a milestone since it's a regression (which is arguably worse than features which just aren't ported yet). Thanks!

lsh-0:
hey, thanks for your attention on this. I've been using Fabric for so many years now I'm trialing Invoke on a personal project to unblock some work projects that need the Python3 upgrade as it doesn't look like Fabric will get there, which is frustrating, but I do appreciate your efforts and those of the other contributors.

bitprophet:
@lsh-0 FWIW, Fabric 2 is technically out now (though it needs a few more feature releases to get up to par with 1.x) and supports Python 3. But it's also built heavily on top of Invoke so working with plain Invoke is a fine place to start; can always use fabric 2's API directly or switch later.

ewanjones:
For future reference, I found a workaround:

```python
def decorator(fn):
    @functools.wraps(fn)
    def _decorated(ctx, *args, **kwargs):
        return fn(ctx, *args, **kwargs)
    return _decorated

@invoke.task
def some_task(ctx, config="Something"):
    @decorator
    def _inner(ctx):
        _run_python_tests(ctx, config)

    return _inner(ctx)
```

This allows `invoke.task` to find the task and inspect the args

corentinbettiol:
[...]

PR #606, opened 2019-01-06
fix #357 - ValueError for annotated tasks

toejough:
As mentioned in #357 - using `getfullargspec` as recommended in the error itself.

---
This change is [Reviewable](https://reviewable.io/reviews/pyinvoke/invoke/606)

PR #666, opened 2019-09-13
Bugfix/issue 555 play nice with other decorator

nhtua:
I made these changes to implement for:
https://github.com/pyinvoke/invoke/pull/246
https://github.com/pyinvoke/invoke/pull/399
https://github.com/pyinvoke/invoke/issues/555

It's more than 4 years from the date when people found it hard to apply to Invoke `@task` on other `@decorator`. The root cause of it is `inspect.getargspec` which couldn't handle decorated function. It's also recommended replacing by `inspect.signature` for a "more structured introspection API for callables" from [python 3.4](https://docs.python.org/3.4/library/inspect.html#inspect.getargspec).

I also keep the compatibility with Python 2.x by using the original function under the name `argspec_python2`.  New `argspec` will check if it is Python 2 and switch the flow to `argspec_python2` on runtime.

### DIFF
--- a/invoke/tasks.py
+++ b/invoke/tasks.py
@@ -145,15 +145,20 @@ class Task(object):
           is a valid value on its own).
 
         .. versionadded:: 1.0
+        .. versionmodified:: 1.3.0
+        The original method (version 1.0) was using inspect.getargspec()
+        which makes problem when get args from decorated function.
+        It is recommended to replace by inspect.signature() from Python 3.0
+        https://docs.python.org/3.6/library/inspect.html#inspect.getargspec
         """
-        # Handle callable-but-not-function objects
-        # TODO: __call__ exhibits the 'self' arg; do we manually nix 1st result
-        # in argspec, or is there a way to get the "really callable" spec?
-        func = body if isinstance(body, types.FunctionType) else body.__call__
-        spec = inspect.getargspec(func)
-        arg_names = spec.args[:]
-        matched_args = [reversed(x) for x in [spec.args, spec.defaults or []]]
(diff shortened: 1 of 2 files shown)


nhtua:
Don't know if we should drop support for Python 2.7 now. It's up to you bitprophet.

nhtua:
I will update some test-cases to ensure `argspec()` will work well with multi decorators.

nhtua:
I've added unit test to cover cases of applying multi decorators (both function and class decorators)

nhtua:
Sorry, Just accidentally clicked the button Request Review. I cancelled the request. Don't want to be rude to ask you review it now. Take your time. Thank you!

lyrixx:
Hello @bitprophet, did you see this PR ? I would be fantastic to have support for decorator :heart_eyes:

phelian:
Another voice for getting this in @bitprophet

hjoukl:
I'd also like to see this go in :-).

FWIW if Python 2.7 compatibility is of prime concern here one could still fall back to use `inspect.getargspec(f.__wrapped__)` for the Python 2.7 case (as you probably know). Provided you can use a modified/enhanced alternative to `functools.wraps` for the decoration, as Py2-`functools.wraps` does not set this convenient attribute.

Some helpful discussion (imho) can be found here: https://stackoverflow.com/a/43506509/7379507

neozenith:
Looks like this overlaps with some other issues reported about type checking in #606 and #458 
I'll have a look at reviewing this area of the code over coming weeks especially if we can get a 2-for-1

bitprophet:
2.0.0 will be out soon & is now using `inspect.signature`!

PR #670, opened 2019-09-23
Fix unreliable local runner on Windows

mrh1997:
The Local runner does not work reliable on Windows. This Problem can be easily reproduced by the following task:

```
@task(default=True)
def reproduce_bug(c):
    while True:
        c.run('git branch', hide=True)
```

At least on my Windows 10 machine (Win Pro 1903 64bit) it  raises an UnexpectedExit exception with exitcode None after 1-10 iterations of the loop. I tested with Python 2.7 (32bit) and 3.7 (64bit).


btorellALTA:
Thanks for this PR, this worked for me. I was having an issue where invoke would exit with an UnexpectedExit while performing `docker build` commands, though it was inconsistent when it would actually exit (sometimes it would exit after running three builds, sometimes after four, sometimes after one). Applying this PR fixed the issue for me.

PR #681, opened 2019-11-22
Allow to use a pathlib.Path instance as a `cd` argument

mariocesar:
``` python
from pathlib import Path

basedir = Path(__file__).parent.resolve()

@task
def deploy(ctx):
    with ctx.cd(basedir):
        ctx.run("ls -al")
```

Will fail with  

```
File "/python/lib/python3.7/site-packages/invoke/context.py", line 94, in run
    return self._run(runner, command, **kwargs)
  File "/python/lib/python3.7/site-packages/invoke/context.py", line 100, in _run
    command = self._prefix_commands(command)
  File "/python/lib/python3.7/site-packages/invoke/context.py", line 242, in _prefix_commands
    current_directory = self.cwd
  File "/python/lib/python3.7/site-packages/invoke/config.py", line 123, in __getattr__
    raise AttributeError(err)
AttributeError: No attribute or config key found for 'cwd'
```

This is due that when getting the `cwd` value it expects to have an string type value.

By converting any cd argument to an string, it now supports to use pathlib.Path instances.

### DIFF
--- a/invoke/context.py
+++ b/invoke/context.py
@@ -366,7 +366,7 @@ class Context(DataProxy):
 
         .. versionadded:: 1.0
         """
-        self.command_cwds.append(path)
+        self.command_cwds.append(str(path))
         try:
             yield
         finally:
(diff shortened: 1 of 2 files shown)


russkel:
Relevant issue #454

bitprophet:
Rolling into #607

PR #691, opened 2020-01-05
Fix wrong assert check

pghmcfc:
The assert_has_call method does not exist but its use did not cause an
issue on older versions of unittest because it was created on demand.

More recent versions of unittest has catch this type of error:
https://bugs.python.org/issue21238

The fix is to use the correct method, which in this case is assert_any_call.


bitprophet:
Nice catch, thanks! Cherry picked to an older release line & merging up.

PR #703, opened 2020-02-22
Update config.py

detlefla:
Current versions of PyYAML display a YAMLLoadWarning, indicating that using the default loader is unsafe. Trying to stay compatible with old versions that might not offer the save_load method.

### DIFF
--- a/invoke/config.py
+++ b/invoke/config.py
@@ -905,6 +905,9 @@ class Config(DataProxy):
 
     def _load_yaml(self, path):
         with open(path) as fd:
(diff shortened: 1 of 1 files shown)


vincentbernat:
`safe_load` is quite old. I don't think you even need to test for existence.

neozenith:
This has been addressed with v1.7.0 where an updated version of PyYAML has been vendored.
https://www.pyinvoke.org/changelog.html

Inparticiular #803

Issue #15, opened 2012-07-06
Logging & other output concerns


bitprophet:
Turning this into a general "how to do logging and useful output controls for invoke and fabric 2" ticket.

## Use cases

- Display the task or tasks being invoked so users can follow the "flow" of nontrivial execution
  - plus parameters
  - plus dependency chain info
    - even skipped/satisfied dependencies
- Display subprocess commands being called so users can tell what's actually being done on their behalf
  - plus which task is currently in play (optionally) in case the per-task info has already scrolled by
- Display some sort of 'prefix', on a line-by-line basis, prepended to subprocess stdout/err - such as the command, task or task parameter that spawned it
  - Command itself less likely (esp since most command strings tend to be long) but task or task param is useful for things like parallel execution (#63) or Fabric per-host exec.
- Probably more besides...
- Must be able to turn a lot of this off or on at will
- May require an entirely different stdout/err mirroring/display mode when prefixes are enabled, since otherwise they'd get in the way of bytewise output (think Unicode decoding, interactive sessions, etc)

## Older thoughts

Thoughts for now, jumping off of the description of fabric/fabric#57 & current state of things in Invoke master:
- Re: log lib to use:
  - May not do anything besides stdlib logging after all, it's still the most-used option, and the core problem of "what to log, when, and how" is not going to change drastically even if we use another lib. It's Not That Bad, Really™.
  - [Structlog](https://structlog.readthedocs.org/) came around after that original ticket started, and actually looks like the best option if we want to add anything on top of stdlib (in part as it's the only major one that supports 2.6-3.x, has lots of good features, actually runs on top of stdlib or whatever you want, and I am mildly biased as I know the author :3)
  - Twiggy is 2.x only, so nope.
  - Logbook is 2.7-3.x, so until we're comfortable dropping 2.6 (not for a while...) it's also a no-go. Still a decent looking option otherwise.
- Log a lot of things by default, mostly at DEBUG for, well debugging. E.g. logic paths used, etc - good for troubleshooting.
  - Invoke does this already; it simply offers `invoke.utils.debug` (w/ setup to add other levels easily) that logs to a package-level logger. Could probably be expanded some, it's mostly used in the parsing and config modules.
- Log 'operational' level things as INFO or similar, sans any 'large' fields - i.e. log when `run` is called, on which host (in Fabric), with which command (tho this may want truncating...), run-time, etc - but _not_ the stdout/stderr contents (though sizes of those, similar to HTTP log formats, is probably good).
[...]

PR #706, opened 2020-02-27
add --verbose flag to indicate INFO logging level

anarcat:
This is useful for fabfiles that use `logging.info` as an intermediate
between 'no output' and 'full debug'.

This is not ideal: I don't like that if/elif check to set the log
level. But I cannot figure out how to do the equivalent of this
argparse snippet:

    parser.add_argument('--verbose', '-v', dest='log_level',
                        action='store_const', const='info', default='warning')
    parser.add_argument('--debug', '-d', dest='log_level',
                        action='store_const', const='debug', default='warning')

I suspect I might be missing some invoke.yml configuration bits as well.

See: #15


anarcat:
i force-pushed in the hope tests now pass.

anarcat:
i've rebased this, again. the remaining CI failures here are test coverage problems. i looked at how `enable_logging` is currently tested and yes, this introduces a new branch that's untested, but looking at `debug_flag_activates_logging` I don't see how this could be tested unless (or until?) `verbose()` calls (similar to the current `debug()` calls) would be added somewhere in the codebase, which is not really the goal here.

the goal here is just to allow the logging level to be set from the outside. i could make a test that checks if the log level is set properly with `-v`, but i'm not sure it's worth the headdesk'ing unless i'm told it's actually required here (and that this patch is actually somewhat interesting upstream as well...)

in the meantime, I've hacked my `fabfile.py` to add this:

```
logging.basicConfig(level=os.environ.get("FABRIC_LOGLEVEL", "INFO"), format="%(levelname)s: %(message)s")
# override default logging policies in submodules
#
# without this, we get debugging info from paramiko with --verbose
for mod in 'fabric', 'paramiko', 'invoke':
    logging.getLogger(mod).setLevel('WARNING')
```

the `FABRIC_LOGLEVEL` is inspired by `INVOKE_DEBUG` but otherwise a complete fabrication here. it's to allow callers to change the loglevel without having to patch fabric/invoke. :)

let me know if you find any of this useful, in any case.

PR #730, opened 2020-05-07
feat: Add ability to source task arg from ealier call

tomdottom:
```python
from invoke import task, call
from invoke.tasks import NOT_SET, FROM_PARENT


@task
def bar(ctx, baz=1):
    print(baz)


@task(pre=[call(bar, baz=FROM_PARENT)])
def foo(ctx, baz=1):
    ctx.run("echo Hello")
```

```shell
$ invoke foo
1
Hello

$ invoke foo --baz=99
99
Hello
```

### DIFF
--- a/invoke/executor.py
+++ b/invoke/executor.py
@@ -1,9 +1,30 @@
-from .util import six
+from collections import defaultdict
+from functools import reduce, wraps
+import inspect
 
 from .config import Config
 from .parser import ParserContext
-from .util import debug
-from .tasks import Call, Task
+from .util import debug, six
+from .tasks import Call, Task, NOT_SET, FROM_PARENT
+
+
+def unwrap(f):
+    while hasattr(f, '__wrapped__'):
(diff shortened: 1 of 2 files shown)


danieleteti:
Very Needed! Thanks!

bossjones:
This would be a cool feature to have, I love pyinvoke ( and fabric ), would be great to pass some of these params around.

AMMullan:
Is there any update on this one? Being able to pass parent task parameters into pre/post tasks is paramount in so many situations and something has limited what we can do with Invoke to date.

josix:
This is really a helpful feature to make shell related lint/reformat workflow more convenient, really nice. Thank you.

blakewaud:
I know volunteers to manage the issue list are limited, but I was looking specifically for how to do this yesterday and spent most of the day trying to come up with a solution. This or something like it is sorely needed!

PR #751, opened 2020-08-17
Don't fail on empty docstrings

matt-hayden:
Hello,

Invoke v1.5.0 on Python 2.7. Methods defined without a docstring, or with an empty docstring, cause an exception at `.splitlines()[0]`. This doesn't cause a problem running the subcommands themselves, but `--list` fails.

Example: run `fab --list` with this:
```
import fabric

@fabric.tasks.task
def foo(ctx):
    """
    """
    print("noop")
```

### DIFF
--- a/invoke/util.py
+++ b/invoke/util.py
@@ -169,7 +169,8 @@ def helpline(obj):
     docstring = obj.__doc__
     if not docstring or docstring == type(obj).__doc__:
         return None
-    return docstring.lstrip().splitlines()[0]
+    lines =  docstring.lstrip().splitlines()
+    return lines[0] if lines else ""
 
 
 class ExceptionHandlingThread(threading.Thread):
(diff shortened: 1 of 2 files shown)


matt-hayden:
@bitprophet , is this worth keeping open?

invoke/util.py
```diff
@@ -169,7 +169,8 @@ def helpline(obj):
     docstring = obj.__doc__
     if not docstring or docstring == type(obj).__doc__:
         return None
-    return docstring.lstrip().splitlines()[0]
+    lines = docstring.lstrip().splitlines()
+    return lines[0] if lines else ""
```
jefftriplett, line 173: @bitprophet would you prefer this return `None`?

jefftriplett (commented):
I will defer to @bitprophet,  but it seems like a reasonable edge-case.

bitprophet:
Thanks for noticing this! I ended up making my own commit:

- IMO the "right" fix here is to try a strip() on the docstring in the early test; this lets us always return either None-or-string (as Jeff T noted). Bit less code, bit cleaner/more obvious.
- The docstring of this method (ironically) was inaccurate re: that issue (None vs empty string) so I also updated it
- Changelog entry
- Committed to release branch and merged to main branch

PR #767, opened 2020-11-25
Implement poetry

kuwv:
This pull request is intended to start the conversation to migrate invoke to Poetry.

Poetry is a modern package manager for Python. Implementing it will help improve project management, dependency management, and is just a more enjoyable than setuptools.

Some additional tasks still need to be done:
- Update CI pipelines to use Poetry
- Update Invocations
- Tests builds


kuwv:
Hi @bitprophet,

Before I continue further, can you tell me your thoughts? I know this is something you wanted to get started. One other question is what is the purpose of - what I'm guessing is a second virtualenv - in the tests?

-Jesse

bitprophet:
Funny timing, I did take a big stab at this a month or so ago (including some of what you've done here, like extracting stuff out of setup.cfg into eg .flake8). I started on a baby project of mine (Lexicon) and it went "okay" but when I tried to apply to Invoke, it was complex enough that I tabled it for the time being (as other tasks, like moving to CircleCI and getting features out the door, are higher priority).

I didn't get all the way through the various things our CI and setup.py do but when I left off:

- Poetry is less amenable to the idea of an "editable, prerelease install" than pip was (ye olde `pip install git+https://github/url#egg=project` to get a checked out latest-main-branch copy of the package).
  - This is less important now in the "users easily installing unreleased versions" (in fact I've already removed mention of it from our docs)
  - but I think the CI code still wants to do this in places - for example making sure that development Invoke and development Invocations grab each other for CI testing of in-flight changes
  - There's also the question of how to do it in a human dev environment - last I looked Poetry still did not have a good story to solve the pip case of "project needs dep X, I have a checkout of X locally, I `pip install -e /path/to/X`, and pip considers the dependency satisfied". This is arguably the same as the CI need.
- Setup.py and project code both using the same version number information.
  - It looks like the "right" solution for this today is to use importlib-metadata. That's a new dependency but otherwise I don't think it's a big problem. I just haven't tried it out yet.
- Python 2 vs 3
  - IIRC this is primarily a problem because we have to bundle discrete codebases for PyYAML on either interpreter. I haven't looked at this in years so it's possible they merged, but it's more likely they just dropped 2.
  - The upshot of that is that we generate 2 wheels, one for Py2, one for Py3
  - Our CI attempts to do the "make sure this version of the code can still be published & installed w/o error" thing, which thus requires that dual wheel building and installing, which in turn requires access to both interpreter families within the same test run (including virtualenvs or at least envs our user can install the 'wheel' package into).
  - This is all irritatingly byzantine and given I'd like to drop Python 2 in 2021 sometime - maybe early! - this is another reason I've tabled some of this work. This stuff is massively simplified once we drop Py2.
  - I'm also likely to do that simplification as I move to CircleCI.

Lotta words, sorry. And I haven't even gone 100% thru setup.py to look for other things Poetry might block. If you've got the energy to take a stab and map out what needs doing or try solutions, feel free.

I would focus on getting things working locally and avoid anything Travis specific - maybe use its .yml as a guide for what tasks our CI does in general, but don't worry about making it actually function on there. Again, I plan to set up orbs n such on Circle (either using, or replacing, a copy of the invocations.travis module).

bitprophet:
Another thing I just ran into (had some work open from a while ago) is that Fabric (the project Invoke split out of) has even MORE nasty but required stuff in its setup.py - even harder to do without in a declarative toml world. Eg it needs to be capable of installing its package and entrypoint CLI stub as different names depending.

There may be solutions to that as well but they're nontrivial and need time to investigate. And I'd prefer to get all my major projects (Invoke, Fabric, Paramiko) updated around the same time so they're not using vastly different packaging tooling.

I'm still hoping to get there eventually (I assume someday pypi or pip will stop supporting setup.py entirely) so if none of this puts you off, any research done is valuable.

kuwv:
Along with Python2 deprecation, yaml vendoring, project coordination, and CI updates I think this would probably be suited as a 2.0 feature then.

As for the versioning, importlib-metadata shouldn't be a problem. I have been using bump2version but have been considering the more modern python-semantic-release.

kuwv:
@bitprophet I implemented importlib-metadata. Now it looks like invocations is needs updates.

I created a merge request to add poetry for invocations there. Or would you like to just update the CI file here first?

kuwv:
@bitprophet

I have been using flit for editable installs. It supports both symlink and pth files.

neozenith:
For what it's worth `poetry>=1.2` is almost out of `pre-release` stage and supporting editable installs (finally).

Good progress has been made to the CircleCI migration which looks like it is currently setup to only test Python 3.6 -> 3.9 at this stage.

Which is in preparation for removing the Python 2 code.

So this might be worth revisiting after the above events have had time to tick over.

bitprophet:
Yea I'm still gently interested in giving this another go, but def after Python 2 removal / Invoke 2.0. (Which should be mostly-just-py2-removal with a few other minor backwards incompat updates - not a big shakeup otherwise.)

PR #769, opened 2020-11-26
Made UnexpectedExit report to stderr

massey101:
It's rather unnerving to have a task exit with failure but not report
anything. If it's okay for the error to be reported when hidden is
enabled it should be okay when it's not enabled.


SergiyKolesnikov:
This feature would be really useful in situations when it is not immediately obvious from then output that some command in a task (or a per-dependency) failed. For example, `make` prints a one-liner, something like
```
make: *** [Makefile:2: foo] Error 2
```

PR #776, opened 2021-01-03
Fix typo

benbridts:


jefftriplett (approved):

jefftriplett:
@benbridts thank you for your contribution!

bitprophet:
I think this was actually on purpose - a bit of irony. But if it's confusing enough to result in a PR then probably not worth the humor!

PR #791, opened 2021-04-26
Adding support for a custom output format via run.echo

davidjmemmett:
The format of the output of fabric2 doesn't quite do it for me; rather than forcing a change that makes it different for everyone, I thought allowing users to customise the output format might be a good idea. It currently only supports wrapping the `command` passed to it, but I will happily look into ways of determining things like whether it's `sudo` or not (and maybe caching the current unadulterated command somewhere to display prettily like fabric1).

- [x] Added supporting test(s)
- [x] Added supporting documentation

Please let me know if there's anything else I've missed.

Cheers,
David

### DIFF
--- a/tests/config.py
+++ b/tests/config.py
@@ -104,6 +104,7 @@ class Config_:
                     "hide": None,
                     "in_stream": None,
                     "out_stream": None,
+                    "output_format": "\033[1;37m{command}\033[0m\n",
                     "pty": False,
                     "replace_env": False,
                     "shell": "/bin/bash",
--- a/invoke/config.py
+++ b/invoke/config.py
@@ -481,6 +481,7 @@ class Config(DataProxy):
                 "hide": None,
                 "in_stream": None,
                 "out_stream": None,
+                "output_format": "\033[1;37m{command}\033[0m\n",
                 "pty": False,
                 "replace_env": False,
                 "shell": shell,
--- a/invoke/runners.py
+++ b/invoke/runners.py
@@ -282,6 +282,16 @@ class Runner(object):
             should be written. If ``None`` (the default), ``sys.stdout`` will
             be used.
 
+        :param output_format:
+            A string, which when passed to Python's inbuilt ``.format`` method,
+            will change the format of the output when ``run.echo`` is set to
+            true.
+            
+            Currently, only ``{command}`` is supported as a parameter.
+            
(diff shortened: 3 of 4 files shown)


bitprophet:
Thanks for this! Made a few changes, yell if you disagree, not sure when I will release.

- Changed to `echo_format` to be a bit clearer, `output_format` sounded like it refers to stdout/err
- It's not clear why you changed from print() to sys.stdout.write plus a newline in the new option.
  - Unless there's a strong upside to that which I missed, I'd rather keep it how it was, you never know who might've been mocking or overriding print() 🙃
  - but mostly, requiring a newline in the option is (IMHO) going to trip people up way more often than it would help somebody with the unusual case of "not wanting a newline at all" (is that a thing?)
- Added changelog entry
- Blackened (CI would've caught this if it was fixed yet. next thing on my agenda after diggin out my inbox...)

tests/config.py
```diff
@@ -104,6 +104,7 @@ def basic_settings(self):
                     "hide": None,
                     "in_stream": None,
                     "out_stream": None,
+                    "output_format": "\033[1;37m{command}\033[0m\n",
```
jcw, line 107: Triggered by the change in 1.6.0, I'd like to point out that this default is not very convenient for terminals with a white background. I constantly need to select the output before I can see what it is ... (MacOS, iTerm2) - please consider choosing anything but white (or black) as default.

bitprophet: To be fair this didn't truly change, and it's now changeable. It's hard to come up with defaults for this kind of thing that work equally well in the two broad color scheme options out there :( eg any sufficiently off-white gray that would be visible on a white BG becomes distressingly low-contrast on a black background.

In my personal travels I see far more developers with black backgrounds than white so it's always seemed the slightly safer option. And users can now change it as needed thanks to the feature-add.

I suspect in the future if we grow additional colorized bits we'd extend it to have a "color scheme" config group so at least then one could package useful defaults for both and you'd only need to tweak one setting saying "light bg". FWIW.

jcw: True (sorry, should have been more specific: the change caused me to look into the underlying code, been there all along). Light blue or light green, perhaps? Even the slightest hue would help. This problem is not specific to pyinvoke, clearly. I just wanted to raise a little flag, no more. Feel free to ignore.

PR #842, opened 2022-01-21
Circle-CI PR test

bitprophet:
Mostly for checks.


bitprophet:
May as well stick some notes here...

* Dropping PyPy for testing, for now, as I cannot find a PyPy circle-ci image easily. I note their images use pyenv so it's probably feasible somehow, later.
* Can't for the life of me remember exactly why I needed to run my test suite _inside sudo_ on Travis; the regular and integration test suites seem to pass OK on Circle without it. IIRC had to do with the default Travis user not being sudo-capable or some such, whereas Circle is running in containers?

PR #843, opened 2022-01-31
Update vendored six library to version 1.16.0

Tenzer:
This version has Python 3.10 a compatibility fix that means the following warning won't be emitted anymore when Invoke is used on Python 3.10:

> ImportWarning: _SixMetaPathImporter.find_spec() not found; falling back to find_module()

See the issue reported for more details: https://github.com/benjaminp/six/issues/341.


Tenzer:
Six was removed in 2ec488b3c201127dd5e5c5e1460135fdee7ecf83 so this is no longer necessary.

PR #849, opened 2022-03-03
[sudo] This small update fix so many situations ...

albundy83:


albundy83:
Could be done with:
```python
cmd_str = "sudo -S -p '{}' {} sh -c \"{}\"".format(prompt, user_flags, command)
```

Varriount:
This assumes `bash` is present on the system (which it might not be!).

albundy83:
Yes you are rigth but sudo command is limited, that's a pity :(

neozenith:
Is there an issue with passing `bash -c "yourcommand"` or `sh -c "yourcommand"` as the command  to the original code here?

PR #864, opened 2022-06-14
Add the option to request a login shell from the context.sudo helper.

unux:


Eidansoft:
In order to use this change from fabric2, it would be needed also a PR into fabric2 repo to make it accept the login_shell param into SerialGroup and ThreadingGroup sudo calls. Is it already created somewhere?? I cannot find it.
Thanks

Issue #782, opened 2021-01-14
Config from imported task does not get merge recursively


alexandre-perrin:
I'm trying to get hierarchical configs to be merged recursively at task call, but it merges only the first level of dictionary. I wonder if this is the expected behaviour.

# Reproduce

```
# project.py
from invoke import task, Collection
from invoke.config import DataProxy
import project

def print_conf(conf, indent=""):
    for k, v in conf.items():
        if isinstance(v, (DataProxy, dict)):
            print(f"{indent}{k}:")
            print_conf(v, indent + "  ")
        else:
            print(f"{indent}{k}: {v}")
@task
def info(c):
    print("project.info.data:")
    print_conf(c.config.data, "  ")

ns = Collection(info)
ns.configure({
    "data": {
        "setting_1": "default_collection",
        "setting_2": "default_collection",
        "setting_3": "default_collection"
    }
})
```

```
# tasks.py
from invoke import task, Collection
import project

@task
def info(c):
    print("info.data:")
    project.print_conf(c.data, "  ")

ns = Collection(info, project)
ns.configure({
    "data": {
        "setting_2": "root",
        "setting_3": "root"
    }
})
```

```
$ inv info project.info
info.data:
  setting_2: root
  setting_3: root
project.info.data:
  setting_2: root
  setting_3: root
```

# Expected

I would expect `project.info.data` to be:

```
  setting_1: default_collection
  setting_2: root
  setting_3: root
```

# Workaround

My current workaround is to monkey patch `_task_with_merged_config` to call `merge_dicts` instead of native `dict` merge:

```
from invoke.collection import merge_dicts

def _task_with_merged_config(self, coll, rest, ours):
    t, c = self.collections[coll].task_with_config(rest)
    return t, merge_dicts(c, ours)

Collection._task_with_merged_config = _task_with_merged_config
```

PR #889, opened 2022-11-12
Fix nested merge in _task_with_merged_config

alexandre-perrin:
Currently, `_task_with_merged_config` returns a single level merged
dict.
Howerver, if the config contains nested dicts, only the first
level gets merged, removing eventual fields non defined
in origin config.

This aims to recursively merge config when retrieving the task config,
and to unify how collection configs are merged.

### DIFF
--- a/invoke/collection.py
+++ b/invoke/collection.py
@@ -363,7 +363,7 @@ class Collection(object):
 
     def _task_with_merged_config(self, coll, rest, ours):
         task, config = self.collections[coll].task_with_config(rest)
-        return task, dict(config, **ours)
+        return task, merge_dicts(config, ours)
 
     def task_with_config(self, name):
         """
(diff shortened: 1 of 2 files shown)


alexandre-perrin:
Fixes https://github.com/pyinvoke/invoke/issues/782

dustinlennon:
I independently came to the same resolution as @alexandre-perrin  -- using `merge_dicts` felt more intuitive to me for an operation described as a recursive merge.  

It also looks like @bitprophet at least considered the proposed changes to [#782](https://github.com/pyinvoke/invoke/issues/782).  However, the PR doesn't seem to have persisted into the latest tagged version (3.0.3); c.f. [invoke/collection.py](https://github.com/pyinvoke/invoke/blob/3.0.3/invoke/collection.py#L374).  

Just wondering if this was intentional and/or if there was a discussion about the rationale here.

Also, I consider myself only barely proficient with git / github, so it's very possible that I've overlooked something obvious.  If so, apologies.

Thanks for any insights here!

Dustin

PR #893, opened 2022-11-29
Drop mock requirement on Python > 3.3

jelly:
Since Python 3.3 mock is also provided by `unittest.mock` which is a buildin preferred over the standalone module.


jelly:
Fixed in main now, closing.

PR #902, opened 2023-01-15
refactor: remove additional python2 cruft

kuwv:
Remove python2 left over cruft from repo.

### DIFF
--- a/invoke/collection.py
+++ b/invoke/collection.py
@@ -8,7 +8,7 @@ from .parser import Context as ParserContext
 from .tasks import Task
 
 
-class Collection(object):
+class Collection:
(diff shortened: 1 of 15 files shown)


kuwv:
@bitprophet I'd recommend doing this cleanup also. This just removes redundant object inheritance stuff that python3 does natively. Just some housekeeping.

bitprophet:
Nice, don't think object inheritance and new-super made it into my overall list for some reason but it fits well with this stage of cleanup. Thanks!

kuwv:
Yeah, I missed it too. Not sure why I didn't notice it earlier.

Issue #901, opened 2023-01-11
Fix Nose deprecation warning with Pytest


kuwv:
Pytest had builtin support for nose that is not deprecated in 7.0. The current tests used by invoke utitlize a `setup()` method which Pytest marks as Nose compatibility. Changing this method to `setup_method()` allows the tests to succeed without warnings.

kuwv:
https://github.com/pyinvoke/invoke/pull/903

bitprophet:
Right, and I think I eventually updated pytest-relaxed to account for `setup_method` style setups too. If I'm right about that, this shouldn't be a problem. Let's roll this into the PR.

PR #903, opened 2023-01-15
tests: refactor to resolve deprecated pytest features

kuwv:
Just following the pytest recommendations here. The tests still function correctly without all the false warnings.

https://github.com/pyinvoke/invoke/issues/901


kuwv:
@bitprophet Do you see any issue merging this?

kuwv:
@bitprophet seems this got reverted along with #906

Issue #376, opened 2016-07-14
Error on passing task to other task


Lense:
``` python
from invoke import task

@task
def preTask(ctx):
    print("preTask")

@task
def testTask(ctx, otherTask=preTask):
    print("otherTask @ ", otherTask)
```

spits out

```
Traceback (most recent call last):
  File "/usr/bin/inv", line 11, in <module>
    sys.exit(program.run())
  File "/usr/lib/python2.7/site-packages/invoke/program.py", line 269, in run
    self._parse(argv)
  File "/usr/lib/python2.7/site-packages/invoke/program.py", line 329, in _parse
    self.parse_tasks()
  File "/usr/lib/python2.7/site-packages/invoke/program.py", line 490, in parse_tasks
    self.parser = Parser(contexts=self.collection.to_contexts())
  File "/usr/lib/python2.7/site-packages/invoke/collection.py", line 325, in to_contexts
    name=primary, aliases=aliases, args=task.get_arguments()
  File "/usr/lib/python2.7/site-packages/invoke/tasks.py", line 205, in get_arguments
    new_arg = Argument(**self.arg_opts(name, default, taken_names))
  File "/usr/lib/python2.7/site-packages/invoke/tasks.py", line 180, in arg_opts
    if default not in (None, NO_DEFAULT):
  File "/usr/lib/python2.7/site-packages/invoke/tasks.py", line 83, in __eq__
    if self.name != other.name:
AttributeError: 'NoneType' object has no attribute 'name'
```

I was originally doing something really dumb because I didn't understand how invoke worked, but I came across this weird case. Feel free to close this.

bitprophet:
Huh, interesting, I think this is because the value you're passing in the signature of `testTask` is another Task object, and the argument examination stuff obviously didn't anticipate that. Thanks for the report!

Perhaps more deeply, `__eq__` may not be well implemented for Task objects - it should say "nope, not equal" if the other object is `None`. 🤔

jr-minnaar:
Hi, while testing/researching inspect.Signature for #357, I also came across this issue.

I'm still making my way through the codebase and don't yet fully understand why the \__eq\__ method fails when a task is passed as an argument, but I've found two solutions that stops it from failing:

1. Add type checking to \__eq\__
    ```
    if not isinstance(other, (Task, Call)):
        return False
    ```

2.  Return False on None and when 'name' attribute does not exist
    ```
    if other is None:
        return False

    if not getattr(other, 'name', None):
        return False
[...]

PR #906, opened 2023-01-20
test: implement type hints

kuwv:
_[EDITOR'S NOTE TO FUTURE SELF: I specifically requested that @kuwv poke at this so I could have a gander at what the codebase feels like with typing rubbed on it. See #904.]_


dev-requirements.txt
```diff
@@ -19,3 +19,9 @@ black>=22.8,<22.9
 setuptools>56
 # Debuggery
 icecream>=2.1
+# typing
+mypy>=0.942
```
Dreamsorcerer, line 23:
```suggestion
mypy==0.991
```

Better to pin to a specific version and let dependabot bump it. Any new version of mypy could result in new errors appearing and breaking the CI (e.g. because new features reveal errors that were uncaught in previous versions).

dev-requirements.txt
```diff
@@ -19,3 +19,9 @@ black>=22.8,<22.9
...
+mypy-extensions>=0.4.3
+typed-ast>=1.4.3
+types-mock>=0.1.3
+types-PyYAML>=5.4.3
```
Dreamsorcerer, line 27: I don't think any of these are relevant (e.g. mypy-extensions appears to be a dead project for experimental features, types-mock provides types for a library not installed or imported anywhere, etc.).

mypy.ini
```diff
@@ -0,0 +1,4 @@
+[mypy]
+warn_return_any = True
+warn_unused_configs = True
+exclude = (integration|invoke/vendor|site|tests|tasks)
```
Dreamsorcerer, line 4: For a more comprehensive check, I'd suggest copying this config: https://github.com/aio-libs/aiohttp-jinja2/blob/master/.mypy.ini

Then comment out any options that introduce too many errors, with the idea that they could be introduced in later PRs iteratively.

Dreamsorcerer (commented):
@bitprophet If you're happy to add typing support, I'm happy to help with reviews. I've done a lot typing across the aio-libs projects and a few others.

Note that typing can be done even more gradually than this PR by using the config file (e.g. you could essentially add a blacklist of files/directories that should ignore errors, then gradually remove them from the blacklist 1 PR at a time).

dev-requirements.txt
kuwv: This is actually cruft from a previous rebase. The `mypy-extensions` should be `typing-extensions` and `types-mock` isn't needed because types-mock was removed.

Dreamsorcerer: Cool, if you add back in typing-extensions, it generally seem good practice to do it conditionally on the Python version (e.g. if all features you need are in Python 3.8+, then only install it on python_version<3.8).

kuwv: Yeah, I'd agree with you but I'm sure @bitprophet will want to vender it. 

My goal is to later add some additional security with LiteralSring. https://peps.python.org/pep-0675/

mypy.ini
kuwv: There were actually two configs due to that rebase `mypy.ini` and `pyproject.toml`. I reconfigured everything to `pyproject.toml`. As for the additional configs I have no problem with it but I think it's best to get @bitprophet feedback so I'll add it commented out for now.

kuwv:
The @bitprophet @Dreamsorcerer merge is near complete. The typing_extensions is `4.1.1` for compatibility with `3.6` but probably need to limit import to `3.6` and `3.7`. Coverage is percentage is off though.

```
python -m venv .venv
pip install -r dev-requirements.txt
mypy .
```

invoke/completion/complete.py
```diff
@@ -25,13 +37,16 @@ def complete(names, core, initial_context, collection, parser):
         # Gently parse invocation to obtain 'current' context.
         # Use last seen context in case of failure (required for
         # otherwise-invalid partial invocations being completed).
+
+        # FIXME: this seems wonky
+        contexts: Union[List[ParserContext], ParseResult]
```
kuwv, line 42: Not sure if there about how this one could be improved.

invoke/collection.py
```diff
@@ -92,67 +93,72 @@ def __init__(self, *args, **kwargs):
...
+            return (
+                self.name == other.name
+                and self.tasks == other.tasks
+                and self.collections == other.collections
+            )
+        return False
 
-    def __ne__(self, other):
+    def __ne__(self, other: object) -> bool:
         return not self == other
```
Dreamsorcerer, line 145: On a side point, I'm sure this is the default behaviour if you don't define this method..

invoke/collection.py
```diff
@@ -92,67 +93,72 @@ def __init__(self, *args, **kwargs):
...
 
-    def __nonzero__(self):
+    def __nonzero__(self) -> bool:
         return self.__bool__()
```
Dreamsorcerer, line 148: And this is Python 2 only.

invoke/completion/complete.py
Dreamsorcerer: ParseResult appears to subclass list. So, I think it should be tweaked to subclass `List[ParserContext]`, or whatever is needed to make it work at runtime. Then this annotation will work with just `List[ParserContext]`.

Dreamsorcerer (commented):
@bitprophet Let me know if you're happy to go with something like this, and then I'll pull it locally and do a full review.

invoke/collection.py
kuwv: I think it's actually recommended not to be overridden either.

invoke/collection.py
kuwv: yeah, left over cruft.

kuwv:
@bitprophet could you let us know what your thoughts are? I'd like to have @Dreamsorcerer feedback while we got him. Thanks!

bitprophet:
Taking a look now, though my general feeling so far is positive 👍🏻

invoke/collection.py
bitprophet: Going by git blame, this looks like it was put in mostly for Python 2's sake (given how 3 overhauled the equality stuff), and overlooked on the recent Py2 drop. So we can prob just drop this.

invoke/completion/complete.py
bitprophet: My understanding after a brief search is that the actual "right" way to do this is to change `ParseResult` to inherit from [MutableSequence](https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableSequence) instead of from `list`, since "a listy thing that may only hold ParserContext objects can't be substituted for any other list".

Is it actually possible to just say "subclasses `List[ParserContext]`"?

For the time being, it's _probably_ best to leave this line (and the class inheritance) alone since changing it is probably technically backwards incompatible, but that's a very weak assertion on my part -- so I'm open to counterarguments if this "contexts is xyz type" line is incredibly gross somehow.

Dreamsorcerer: > My understanding after a brief search is that the actual "right" way to do this is to change `ParseResult` to inherit from [MutableSequence](https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableSequence) instead of from `list`, since "a listy thing that may only hold ParserContext objects can't be substituted for any other list".

`MutableSequence` is an abc. So, if you implement all of the methods for that abc, you can do that. But, subclassing from `list` is also perfectly valid and possibly beneficial if it gives you several methods for free.

From a typing perspective `list[ParserContext]` is a list that can't be substituted for other lists which might contain other types. In this situation, it doesn't look like MutableSequence is needed to me.

> Is it actually possible to just say "subclasses `List[ParserContext]`"?

Generally an annotation refers to subclasses, so unless I'm misunderstanding your comment, that already means what you want it to mean.

> For the time being, it's _probably_ best to leave this line (and the class inheritance) alone since changing it is technically backwards incompatible (i.e. clean this up next major rel), but that's a very weak assertion on my part -- I'm open to counterarguments if this "contexts is xyz type" line is incredibly gross somehow.

If we were supporting Python 3.9+ only, then we would subclass `list[ParserContext]`, which at runtime is identical to subclassing `list`. To support Python 3.7/8, we need to make use of `typing.List` though. I just tested it at runtime, and subclassing `List` works fine, and `isinstance(custom_list, list)` passes successfully too. So, I think this change is safe and backwards compatible.

Dreamsorcerer:
Alright, I'll pull it locally at the weekend and go through it fully.

Dreamsorcerer:
@bitprophet Just looking at the vendor parts, not sure what your thoughts are, but I'm not too sure about vendoring typing_extensions. It seems to cause extra type errors and should be a pretty straightforward library to use as an external dependency (plus, can probably avoid installing entirely on recent versions of Python).

I'm also confused by this import:
https://github.com/pyinvoke/invoke/blob/main/invoke/parser/context.py#L3-L6

Why would that import ever fail?

bitprophet:
@Dreamsorcerer The odd looking try/except imports are because some downstream packagers (usually Linux distros) have Opinions™ about vendoring, and wanted the code to gracefully deal with situations where they've not distributed the `vendor` subtree.

I consider(ed) this to be in line with the vendoring itself, in a "neither of these solutions is elegant but in tandem they make it a lot easier for this tool to be trivially available as a CLI app for the most users" sense.

bitprophet:
Note to self or @kuwv - before we merge this wants a changelog entry and I assume adding a CI task that runs `mypy invoke/` or similar. Also, that command gives me a few more typing errors when I run it locally - not sure if that means I'm running something differently than you were?

Dreamsorcerer:
> @Dreamsorcerer The odd looking try/except imports are because some downstream packagers (usually Linux distros) have Opinions™ about vendoring, and wanted the code to gracefully deal with situations where they've not distributed the `vendor` subtree, but have instead ensured via the package manager that external versions of those libraries are around.

Right, makes sense. They will be fixed to specific version anyway as part of the distro packaging, so it makes sense for them to use an external package which they know has relevant security fixes etc.

> EDIT: so, re: typing_extensions, my assumption was that it falls under the other vendored packages, but I hadn't yet grokked exactly why it was brought in to begin with.

Why it was imported? That's because some types are not available in older releases, for example `typing.Literal`. If we don't want anything fancy (the newer typing things are more finetuning that anything fundamental), then we can probably have typing_extensions only for Python <= 3.8.

So, you'd end up with a few imports like:
```
if sys.version_info >= (3, 8):
    from typing import Literal
else:
    from typing_extensions import Literal
```
(mypy will understand this correctly with version_info compares, it won't like a try/except).

> Note to self or @kuwv - before we merge this wants a changelog entry and I assume adding a CI task that runs `mypy invoke/` or similar. Also, that command gives me a few more typing errors when I run it locally - not sure if that means I'm running something differently than you were?

I'm playing with the config at the moment, but one of the things I recommend is `files = ...`, then you only need to run `mypy`, no arguments, and you don't need to worry about remembering which paths to use in every place you call it.

invoke/completion/complete.py
kuwv: Yeah, I think I might need to clarify. I already knew `ParseResult` was a list, but kept forgetting about it when I was looking at it. TBC, this seems more like a issue communicating intent.

To me it reads as it should be `List[Union[ParserContext, ParseResult]]` or `Union[List[ParserContext], Parseresults]`.

Hope that helps.

invoke/collection.py
kuwv: roger, I'll take care of it

kuwv:
> Note to self or @kuwv - before we merge this wants a changelog entry and I assume adding a CI task that runs `mypy invoke/` or similar. Also, that command gives me a few more typing errors when I run it locally - not sure if that means I'm running something differently than you were?

I'm running it on Python 3.10. I'll test is with each of the respective versions this weekend. Also, I plan on adding a pytest plugin by @dbader to run mypy with pytest https://pypi.org/project/pytest-mypy/. We can refactor it later but I'd like the CI pipeline to run mypy and this is will work.

invoke/loader.py
```diff
@@ -27,7 +29,7 @@ def __init__(self, config=None):
             config = Config()
         self.config = config
 
-    def find(self, name):
+    def find(self, name: str) -> Tuple[IO, str, Tuple[str, str, int]]:
```
bitprophet, line 32: This might be a good time for us to switch to using `importlib` since `imp` goes away next major Python release? Note to self that we want to dig up whatever PRs are filed for that.

invoke/parser/context.py
```diff
@@ -83,23 +89,23 @@ def __init__(self, name=None, aliases=(), args=()):
         ``for arg in args: self.add_arg(arg)`` after initialization.
         """
         self.args = Lexicon()
-        self.positional_args = []
+        self.positional_args: List[Argument] = []
         self.flags = Lexicon()
-        self.inverse_flags = {}  # No need for Lexicon here
+        self.inverse_flags: Dict[str, str] = {}  # No need for Lexicone
```
bitprophet, line 94: why was the comment changed? :D

invoke/parser/parser.py
```diff
@@ -1,22 +1,43 @@
...
+from ..util import debug  # type: ignore
 
+if TYPE_CHECKING:
+    from .context import ParserContext
 
-def is_flag(value):
-    return value.startswith("-")
 
+def is_flag(value: str) -> bool:
+    return bool(value.startswith("-"))
```
bitprophet, line 19: Doesn't `str.startswith` always return True/False? Why the `bool`? (And ditto for the next, identical change.)

invoke/program.py
```diff
@@ -606,10 +618,10 @@ def called_as(self):
 
         .. versionadded:: 1.2
         """
-        return os.path.basename(self.argv[0])
+        return os.path.basename(self.argv[0]) if self.argv else "invoke"
```
bitprophet, line 621: This looks a lot like changing behavior! And I don't see a counterpart change elsewhere that would explain it, unless I missed one. What's up?

bitprophet (commented):
Leaving my review off partway in case I'm not able to jump right back in tomorrow. Should be a bunch of line notes visible now. Will pick up from runners.py later!

invoke/collection.py
```diff
@@ -92,67 +93,69 @@ def __init__(self, *args, **kwargs):
...
             self._add_object(arg)
         # Dispatch kwargs
         for name, obj in kwargs.items():
             self._add_object(obj, name)
 
-    def _add_object(self, obj, name=None):
+    def _add_object(
+        self, obj: Any, name: Optional[str] = None
+    ) -> Callable[..., Any]:
+        method: Callable[..., Any]
```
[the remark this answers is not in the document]
Dreamsorcerer: Technically, but mypy may complain if you don't put the subscription in (it's probably one of the configurable options).
This is explicit, so it atleast suggests that someone has thought the typing, whereas missing it out entirely suggests someone hasn't thought about it (maybe didn't even realise that it should be subscripted).

invoke/collection.py
```diff
@@ -218,7 +221,9 @@ def instantiate(obj_name=None):
                 ret = instantiate(obj_name=obj.name)
                 ret.tasks = ret._transform_lexicon(obj.tasks)
                 ret.collections = ret._transform_lexicon(obj.collections)
-                ret.default = ret.transform(obj.default)
+                ret.default = (
+                    ret.transform(obj.default) if obj.default else None
```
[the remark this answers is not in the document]
Dreamsorcerer: If it's internal code, not part of the API, I'd probably go for the second option if it results in cleaner code.

If you go for the first option though, you'll want to type as:
```
_NameT = TypeVar("_NameT", str, None)
def transform(self, name: _NameT) -> _NameT:
```

Which means the return type is the same as the parameter.

i.e. `reveal_type(transform("foo"))` would give you `str`. If you used unions, it'd give you `str | None`.

invoke/config.py
```diff
@@ -808,7 +819,9 @@ def load_collection(self, data, merge=True):
         if merge:
             self.merge()
 
-    def set_project_location(self, path):
+    def set_project_location(
+        self, path: Optional[Union[PathLike, str]]
```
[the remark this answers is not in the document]
Dreamsorcerer: Well, this specific example only uses the variable once, in os.path.join(). That function accepts `str | PathLike[str]`, so mypy is correct here as long as you add the `[str]`.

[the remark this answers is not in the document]
Dreamsorcerer: Minor nitpick, but I'd also drop the Optional when you are already using a union and just do:
```suggestion
        self, path: Union[PathLike[str], str, None]
```

invoke/config.py
```diff
@@ -964,7 +979,7 @@ def _merge_file(self, name, desc):
             # the negative? Just a branch here based on 'name'?
             debug("{} not found, skipping".format(desc))
 
-    def clone(self, into=None):
+    def clone(self, into: Optional[Type["Config"]] = None) -> "Config":
```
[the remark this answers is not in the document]
Dreamsorcerer: I've not looked at the implementation, but `Type[]` means the class (not an instance).

i.e. `foo: Type[Config] = Config` or `foo: Config = Config()`

Dreamsorcerer:
> I'm running it on Python 3.10. I'll test is with each of the respective versions this weekend.

Generally not necessary, you can probably just put `python_version = 3.7` in the config and test it will work for older versions.

> Also, I plan on adding a pytest plugin by @dbader to run mypy with pytest https://pypi.org/project/pytest-mypy/. We can refactor it later but I'd like the CI pipeline to run mypy and this is will work.

Not sure what the current CI looks like, but I've had bad experiences with these kind of plugins. It adds complexity, another dependency, and usually causes problems at some point. I'd recommend just adding it as a separate command in a lint job or similar (a Makefile can always be used to help developers run all the correct steps locally).

Dreamsorcerer:
To make typing easier, and probably make the code a bit simpler, I think invoke/terminals.py could be split up. e.g. terminals_win.py and terminals_unix.py, with terminals.py just importing the relevant one. That would avoid all the extra platform checks and if statements through the module.

I'll just type ignore any errors for now, but something that might be worth someone looking at later.

Dreamsorcerer:
Also just noticed that fluidity-sm hasn't been touched in 12 years. I think that's long reached the point where it would be a good idea to release your own fork or make it part of invoke without vendoring. :P

kuwv:
> Also just noticed that fluidity-sm hasn't been touched in 12 years. I think that's long reached the point where it would be a good idea to release your own fork or make it part of invoke without vendoring. :P

@dreamsorcerer I actually have an overhauled fork of fluidity which I initially named fluidstate and added typing to it. I later renamed it superstate after I decided it would be much better for my other work as a statechart. I am nearly finished with it. I plan on submitting a PR to add it later.

invoke/context.py
```diff
@@ -197,8 +214,9 @@ def _sudo(self, runner, command, **kwargs):
         cmd_str = "sudo -S -p '{}' {}{}{}".format(
             prompt, env_flags, user_flags, command
         )
+        # FIXME pattern should be raw string prompt.encode('unicode_escape')
         watcher = FailingResponder(
-            pattern=re.escape(prompt),
+            pattern=re.escape(prompt),  # type: ignore
```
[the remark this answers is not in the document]
kuwv: The return from `re.escape` is a string but `pattern` is now a raw string buffer instead. This type mismatch is a bug.

invoke/completion/complete.py
```diff
@@ -2,16 +2,28 @@
...
+from typing import List
 import glob
 import os
 import re
 import shlex
+from typing import TYPE_CHECKING, Union
 
 from ..exceptions import Exit, ParseError
-from ..util import debug, task_name_sort_key
+from ..util import debug, task_name_sort_key  # type: ignore
```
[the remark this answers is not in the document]
kuwv: This is due to the `debug` function being injected into the globals of the `utils` module. It doesn't propagate types for mypy.

invoke/parser/context.py
kuwv: ugh, I have no excuse

invoke/parser/parser.py
kuwv: This is cruft left over from troubleshooting if these strings were more complex types or plain strings. Cleaning it up now.

invoke/program.py
kuwv: I should have commented on that one instead - slops on my part. The argv is `Optional` and it should probably return `None` instead. I'll add a comment but I think we probably need to dig a little on what should happen when it is None.

invoke/context.py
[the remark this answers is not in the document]
Dreamsorcerer: Actually, it's not. I'll try and push my proposed changes to a branch tomorrow, but I've removed this type ignore.

invoke/completion/complete.py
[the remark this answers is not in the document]
Dreamsorcerer: I didn't see why the `globals()` was needed. This all seems to resolve after changing the code to `debug = log.debug`.

kuwv:
> > I'm running it on Python 3.10. I'll test is with each of the respective versions this weekend.
> 
> Generally not necessary, you can probably just put `python_version = 3.7` in the config and test it will work for older versions.
> 
> > Also, I plan on adding a pytest plugin by @dbader to run mypy with pytest https://pypi.org/project/pytest-mypy/. We can refactor it later but I'd like the CI pipeline to run mypy and this is will work.
> 
> Not sure what the current CI looks like, but I've had bad experiences with these kind of plugins. It adds complexity, another dependency, and usually causes problems at some point. I'd recommend just adding it as a separate command in a lint job or similar (a Makefile can always be used to help developers run all the correct steps locally).

Yeah, really was just looking for a simple way to align the coverage. Yeah, I think I'm going to skip it.

invoke/program.py
kuwv: Actually, I think I modified argv earlier to default to an empty list. Accessing this out of order will throw an index error instead. What would you like to see here if the list is empty?

invoke/collection.py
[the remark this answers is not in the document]
kuwv: Yeah, I'd prefer just plain callable here but it was complaining.

invoke/collection.py
```diff
@@ -92,67 +93,69 @@ def __init__(self, *args, **kwargs):
...
         )
 
-    def __eq__(self, other):
-        return (
-            self.name == other.name
-            and self.tasks == other.tasks
-            and self.collections == other.collections
-        )
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, Collection):
```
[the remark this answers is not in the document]
kuwv: #376

invoke/config.py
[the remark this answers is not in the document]
kuwv: I personally prefer to use only use one way to repres `None` but I know that's not everyone's preference. Combined all `Optinonal` / `Union`.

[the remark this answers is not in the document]
Dreamsorcerer: Well, it's just an alias to `Union[..., None]`, so if you already have a union, it keeps it simpler, rather than nesting expressions.

Regardless, it doesn't matter long term, this will all get changed to `PathLike[str] | str | None` at a later date.

invoke/context.py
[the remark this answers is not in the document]
kuwv: yeah, I was mistaking this for something else. I worked out the bytes/str earlier so this is unneeded. I will add LiteralString later when in a separate PR when we look at `popen` improvements.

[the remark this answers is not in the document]
Dreamsorcerer: Umm, there's no literal string here...

invoke/config.py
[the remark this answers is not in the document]
kuwv: Let me clarify: "There should be one-- and preferably only one --obvious way to do it."

What I mean by "opinion" is that `Union` looks correct for some people but `Optional` qualifies best IMO.

invoke/loader.py
kuwv: @bitprophet I think so too. Could we tackle that after this PR?

invoke/completion/complete.py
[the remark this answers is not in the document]
kuwv: Done

invoke/completion/complete.py
kuwv: Is this resolved?

invoke/collection.py
kuwv: done

invoke/collection.py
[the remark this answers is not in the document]
kuwv: IMO transform doesn't actually do anything when `None` and passing `None` to every other caller would still require the checks. I'd rather throw an exception TBH.

invoke/program.py
kuwv: @bitprophet is the empty sting fine here since it essentially means that function was provided no arguments?

invoke/config.py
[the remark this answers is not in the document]
Dreamsorcerer: _clone_init_kwargs() was wrong.

Dreamsorcerer:
OK, finished going through it. Here are my suggested changes: https://github.com/Dreamsorcerer/invoke/commit/7a9a9ddd14a6e891276962b663ecdd362992dd98

I also started adding tests/ to the checks, but there are a lot off errors occurring (not sure if pytest does something to the classes, but there are lots of nested classes which then access attributes on themselves which only exist in the outer class, which doesn't make much sense).

Typing the tests is one of the places that the public API will get tested, so I already found a few issues with the typing from that. I suggest we come back to tests/ in the next PR.

Dreamsorcerer:
I've also added TODO comments for typing enhancements that can be added in a later Python version, or with typing_extensions (I'll leave it to you to decide whether you want to bring typing_extensions in or not, seems we don't need it currently).

kuwv:
> I've also added TODO comments for typing enhancements that can be added in a later Python version, or with typing_extensions (I'll leave it to you to decide whether you want to bring typing_extensions in again or not, seems we don't need it currently).

Yeah, 

> OK, finished going through it. Here are my suggested changes: [Dreamsorcerer@7a9a9dd](https://github.com/Dreamsorcerer/invoke/commit/7a9a9ddd14a6e891276962b663ecdd362992dd98)
> 
> I also started adding tests/ to the checks, but there are a lot of errors occurring (not sure if pytest does something to the classes, but there are lots of nested classes which then access attributes on themselves which only exist in the outer class, which doesn't make much sense).
> 
> Typing the tests is one of the only places that the public API will get tested, so I already found a few issues with the typing from that. I suggest we come back to tests/ in the next PR.

@Dreamsorcerer I had to actually pin mypy to 0.971 to work with Python3.6.

```
# pip install mypy==0.991
ERROR: Could not find a version that satisfies the requirement mypy==0.991 (from versions: 0.1, 0.11, 0.12, 0.13, 0.14, 0.15, 0.16, 0.17, 0.18, 0.19, 0.20, 0.21, 0.221, 0.222, 0.223, 0.224, 0.225, 0.226, 0.227, 0.228, 0.229, 0.230, 0.231, 0.232, 0.233, 0.234, 0.235, 0.236, 0.237, 0.238, 0.239, 0.240, 0.241, 0.250, 0.251, 0.252, 0.253, 0.254, 0.255, 0.256, 0.470, 0.471, 0.501, 0.510, 0.511, 0.520, 0.521, 0.530, 0.540, 0.550, 0.560, 0.570, 0.580, 0.590, 0.600, 0.610, 0.620, 0.630, 0.641, 0.650, 0.660, 0.670, 0.700, 0.701, 0.710, 0.711, 0.720, 0.730, 0.740, 0.750, 0.760, 0.761, 0.770, 0.780, 0.781, 0.782, 0.790, 0.800, 0.812, 0.900, 0.901, 0.902, 0.910, 0.920, 0.921, 0.930, 0.931, 0.940, 0.941, 0.942, 0.950, 0.960, 0.961, 0.971)
ERROR: No matching distribution found for mypy==0.991
```

@bitprophet Do you want to add typing to tests? It seems that there are some issues with it if we do. Also, can you take a peek at @Dreamsorcerer's work and lmk what you would like to see?

Dreamsorcerer:
> @Dreamsorcerer I had to take your earlier advice and pin mypy to 0.971 to work with Python3.6.

There's no reason to run mypy under Python 3.6. It's a static type checker, so the version of Python that is being run shouldn't affect any results. Just run it in the linter step with a more recent version (I assume you're not using an ancient version of flake8 because they no longer run under 3.6/7), and set the python_version in the config.

As of yesterday, the version can be pinned to 1.0.0.

Dreamsorcerer:
> > I've also added TODO comments for typing enhancements that can be added in a later Python version, or with typing_extensions (I'll leave it to you to decide whether you want to bring typing_extensions in again or not, seems we don't need it currently).
> 
> Yeah, I'll probably add it again later when I rewrite #698.

Sure, but as before, I'd recommend not vendoring it if you bring it in. mypy special cases typing_extensions, and by pulling it into the repo, you lose that special case and require mypy to parse the whole thing, which seems to result in additional errors.

> Do you want to add typing to tests? It seems that there are some issues with it if we do.

The main thing is just figuring out what is going on with those classes and whether it's worth simplifying them to something more intuitive. All the tests seem to be written like this:

```
class Foo:
    def foo(self): ...

    class InnerFoo:
        def test(self):
            self.foo()  # How can this be defined?
```

In normal Python code, that is clearly broken. If those tests are actually working, then pytest or something must be doing something weird to modify the classes.

Dreamsorcerer:
An obvious solution would be to use normal inheritance:
```
class Foo:
    def foo(self): ...

class InnerFoo(Foo):
    def test(self):
        self.foo()
```

kuwv:
> An obvious solution would be to use normal inheritance:
> 
> ```
> 
> class Foo:
> 
>     def foo(self): ...
> 
> 
> 
> class InnerFoo(Foo):
> 
>     def test(self):
> 
>         self.foo()
> 
> ```

I'd like to wait on doing the tests since the tasks and argument need more attention.

Essentially, 'task' is a decorator wrapping another decorator. This is causing Callable => Task => Callable.

Dreamsorcerer:
> I'd like to wait on doing the tests since the tasks and argument need more attention.

That was my suggestion. It'd be too much to change tests in this PR.

> Essentially, 'task' is a decorator wrapping another decorator. This is causing Callable => Task => Callable.

Think this is already done in my suggested changes?

Dreamsorcerer:
Let me know if you want me to put those changes into a PR, so they can reviewed separately.

bitprophet:
Hey @Dreamsorcerer, @kuwv - I'll try to review the rest of this, and the discussion, today. At a high level, +1 on waiting on tests (it's an idiosyncratic pytest plugin that I'm not currently willing to drop, so if it means they cannot _be_ reasonably typechecked, then maybe that's just moot).

----

What exactly is the deal with typing_extensions? Is it something we can specify as part of the development-only dependency list (and perhaps add to some docs in case that's insufficient for folks relying on our type annotations incidentally)?

I like the idea of supporting typing, but not to the point where we'd grow our first non-vendored dependency solely to support typechecking. (Maybe as a setuptools extra, eg `invoke[typed/typing/whatever]`? That might be the best of both worlds?)

kuwv:
> Let me know if you want me to put those changes into a PR, so they can reviewed separately.

I've updated most of your suggestions. I'm looking at the 'task' library though. It's not using 'decorator' lib or 'wraps'. It needs to provide its metadata and the function but not idiomatically.

I'll be traveling and out of pocket here and there.

Dreamsorcerer:
> I'm looking at the 'task' library though. It's not using 'decorator' lib or 'wraps'. It needs to provide its metadata and the wrapped function but it's not doing it idiomatically.

I probably wasn't looking too closely, it just looked like something that wraps a function (similar to asyncio.Task), so I didn't think about wraps(). If that's relevant, feel free to add it in, but not sure it's needed for typing anyway.

Dreamsorcerer:
> What exactly is the deal with typing_extensions?

Pretty much includes newer typing features for backwards compatibility (including some experimental features not yet in a Python release).

Typical use case is to do something like:
```
if sys.version_info >= (3, 8):
    from typing import Literal
else:
    from typing_extensions import Literal
```

Then you can add it as a dependency only for older Python releases, for example in aiohttp we only depend on it for Python 3.7:
https://github.com/aio-libs/aiohttp/blob/3.9/setup.cfg#L56

As I said above, if you don't want to depend on it at all, I've put TODOs in that can be updated when you drop support for the preceding Python version. The TODOs improve the accuracy of typing, but they are not game-changing.

> Is it something we can specify as part of the development-only dependency list (and perhaps add to some docs in case that's insufficient for folks relying on our type annotations incidentally)?

No, it is needed at runtime.

Dreamsorcerer:
> As I said above, if you don't want to depend on it at all, I've put TODOs in that can be updated when you drop support for the preceding Python version.

When I drop support for a Python version, I actually look through the code base with (e.g. when dropping Python 3.6):
```
fgrep -Rn -e '36' -e '3.6' -e '3,6' -e '3, 6'
fgrep -Rn -e '37' -e '3.7' -e '3,7' -e '3, 7'
```
Which helps me to find all the TODOs and version_info checks that can now be removed.

bitprophet:
Yea, I tend to do similar, `# TODO: nuke/change/etc when dropping Python 3.6`, then `ack -i "py(thon)? ?3\.6"` or whatnot.

And more to the point, acknowledged, thanks for the explainer. Given this is about accuracy/gentle degradation of typing ability, and not "its lack makes typechecking impossible or completely pointless", I think I'm fine with the TODOS+eventual updating approach.

I'm also open to changing my approach if I really get the typing religion and/or we get a lot of folks using this initial release of support who then return with "argh the lack of what typing_extensions would enable, is really killing us here!".

bitprophet:
FWIW given both of y'all have already put a lot of time into this and one of you, at least, is partly AFK - I'm likely to make some executive decisions and kick some version of this out the door without a ton more back/forth, so the community can try it out.

Especially since my read of annotations is that we can hone them over time without it being a huge headache for people (vs "real" APIs which can break folks pretty bad if they change unexpectedly). So both gradual extensions of the annotations & occasional tweaking seem doable w/o a lot of hand-wringing. (I do so love hand-wringing...)

invoke/completion/complete.py
bitprophet: It is now, look out for an upcoming push from me. tl;dr Sam was right that we can just subclass `List[ParserContext]` instead of `list`. This simplifies the referenced spot so it no longer requires an awkward Union; it also found a minor typing mistake elsewhere. Hey, the system works!

bitprophet:
Finished reviewing Jesse's changes, made a small pile of my own where I disagreed, and now rolling through Sam's suggestions-via-commit (which I expect to manually apply so it doesn't pull in the test suite changes). Thanks again for all of this!

bitprophet:
Whoops, I see @kuwv doing some of the same thing this evening (pulling in Sam's changes). I'll stop what I'm doing and wait for a go-ahead 😂 

I will note that so far I've liked Sam's changes except for the Generics stuff in invoke.tasks, which I continue to find pretty confusing. While the more (heh) generic use of Any that Jesse started out with is, surely, a lot more broad than the use of the overload decorator + TypeVar/Generic, I find it much easier to read/grok offhand.

Dreamsorcerer:
> I will note that so far I've liked Sam's changes except for the Generics stuff in invoke.tasks, which I continue to find pretty confusing. While the more (heh) generic use of Any that Jesse started out with is, surely, a lot more broad than the use of the overload decorator + TypeVar/Generic, I find it much easier to read/grok offhand.

Let me know if you want to go over anything. `Any` just means dynamic typing, so we lose all typing benefits whenever you see that. So, any function passed into a task becomes untyped, which for such a generic object is rather undesirable.

The `TypeVar` is almost like a placeholder and just means you'll get the same type out as you put in (in this case the `_T` in the body defines the return type that comes from `__call__()`). So, with the `TypeVar` in place this will work:
```
def foo() -> str: ...

t = Task(foo)
result = t()
reveal_type(result)  # str
```

Without that, you'll get `Any`.


The overloads are just for the decorator, which is a little awkward because it can be used as `@task` or `@task()`, so it's just trying to define both cases. Admittedly, the arguments in the signature feel slightly mangled in order to make it work, but again if you don't do that then every single function you put that decorator on is going to become untyped.

Dreamsorcerer:
I think the only thing I'm not entirely certain on is `invoke.parser.argument.Argument`, which is really rather complicated from a static typing perspective, and I'm not sure how precise the types are. Because it has things like value and raw_value, it can have a default value, it can also be None, it might be a list which then append values, and set_value() may cast or not cast; it's all rather complicated as to what the correct types might be...

kuwv:
@bitprophet typing_extensions are the types backport support for Python itself. For example the security related LiteralString would be available to lower versions of python.

kuwv:
@bitprophet I decided I needed to provide my latest updates. Sam fixed a bunch of stuff I overlooked. I'll leave the rest for you to decide. Still need to tune argument and tasks.

bitprophet:
> 

Arguments literally stand in for CLI flags - so like most other such libraries, the possibilities for what type they end up presenting as, are extremely varied. I'm not sure offhand what the "right" way to do this would be (ie if one were to write Invoke from scratch with typing in mind).

However I'm happy kicking the can down the road a bit for the thornier corners here - I assume that just being able to run mypy on client codebases (without having to mark Invoke as entirely untyped, as they would have prior) will be a big step up for many folks, and as I noted upthread, I expect we'll streamline this in the future.

kuwv:
@bitprophet I think this might have gotten reverted.

bitprophet:
Yea, that's messed up. Clearly lazygit is a little too powerful sometimes. Will try to get it back.

bitprophet:
@kuwv Should be restored now. Suspect this was due to me trying to juggle changes between 2.0 and main, must have somehow reset main to be equal to a version of 2.0. The trouble with trying to keep even a single stable branch around for folks…

PR #915, opened 2023-02-12
Add decode_stdin/stdout to skip encoding

### LINKED ISSUE #818: Encoding/decoding binary stream breaks application
I'm trying to stream data with fabric to another server through an SSH connection. The data is just binary data and should have no meaning or anything applied to it within the application. But, when sending this via fabric, invoke will try to encode it as a str and then decode it again as bytes, which breaks everything and results in the socket closing.

My application runs fine after removing those steps from invoke:
```diff
--- a/invoke/runners.py
+++ b/invoke/runners.py
@@ -791,12 +791,6 @@ class Runner(object):
         bytes_ = None
         if ready_for_reading(input_):
             bytes_ = input_.read(bytes_to_read(input_))
-            # Decode if it appears to be binary-type. (From real terminal
-            # streams, usually yes; from file-like objects, often no.)
-            if bytes_ and isinstance(bytes_, six.binary_type):
-                # TODO: will decoding 1 byte at a time break multibyte
-                # character encodings? How to square interactivity with that?
-                bytes_ = self.decode(bytes_)
         return bytes_
 
     def handle_stdin(self, input_, output, echo):
@@ -976,7 +970,7 @@ class Runner(object):
         """
         # Encode always, then request implementing subclass to perform the
         # actual write to subprocess' stdin.
-        self._write_proc_stdin(data.encode(self.encoding))
+        self._write_proc_stdin(data)
 
     def decode(self, data):
         """
```

Dreamsorcerer:
Fixes #818.

### DIFF
--- a/invoke/runners.py
+++ b/invoke/runners.py
@@ -338,6 +338,10 @@ class Runner:
             Which shell binary to use. Default: ``/bin/bash`` (on Unix;
             ``COMSPEC`` or ``cmd.exe`` on Windows.)
 
+        :param bool stdin_bytes:
+            If ``True``, stdin should be treated as raw binary data and will
+            therefore skip any encode/decode steps.
+
         :param timeout:
             Cause the runner to submit an interrupt to the subprocess and raise
             `.CommandTimedOut`, if the command takes longer than ``timeout``
--- a/invoke/runners.py
+++ b/invoke/runners.py
@@ -825,7 +825,7 @@ class Runner:
                     raise
             # Decode if it appears to be binary-type. (From real terminal
             # streams, usually yes; from file-like objects, often no.)
-            if bytes_ and isinstance(bytes_, bytes):
+            if not self.opts["stdin_bytes"] and bytes_ and isinstance(bytes_, bytes):
                 # TODO: will decoding 1 byte at a time break multibyte
                 # character encodings? How to square interactivity with that?
                 bytes_ = self.decode(bytes_)
@@ -1014,7 +1014,8 @@ class Runner:
         """
         # Encode always, then request implementing subclass to perform the
         # actual write to subprocess' stdin.
-        self._write_proc_stdin(data.encode(self.encoding))
+        d = data if self.opts["stdin_bytes"] else data.encode(self.encoding)
+        self._write_proc_stdin(d)
 
     def decode(self, data: bytes) -> str:
         """
--- a/invoke/runners.py
+++ b/invoke/runners.py
@@ -338,8 +338,8 @@ class Runner:
             Which shell binary to use. Default: ``/bin/bash`` (on Unix;
             ``COMSPEC`` or ``cmd.exe`` on Windows.)
 
-        :param bool stdin_bytes:
-            If ``True``, stdin should be treated as raw binary data and will
+        :param bool decode_stdin:
+            If ``False``, stdin should be treated as raw binary data and will
             therefore skip any encode/decode steps.
 
         :param timeout:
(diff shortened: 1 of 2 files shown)


Dreamsorcerer:
@bitprophet Not sure what you pushed, but seems to have added a bunch of unrelated changes to the diff.

Dreamsorcerer:
@bitprophet ?

bitprophet:
Yea looks like the typing updates made github lose its mind. if you can rebase and re-push that'd probably be ideal - you'll probably want to make sure you copy over any type hints from `main` to your branch first as that's likely the culprit.

----

From skimming what looks like the two commits actually relevant: my off the cuff thoughts:

- if we go this route we may want to rename the option to `stdin_is_bytes=True` or maybe `decode_stdin=False`
- however I'm wondering if there's anything cleaner/smarter we can do within the "number of bytes to read" helper instead (per your original 'fix' for this elsewhere)
- or, since the "it's binary, please do not do any edge encoding/decoding" approach is arguably more correct, is there anything smarter we can do to autodetect?
  - are there approaches other tools use when handling unknown stdin?
  - eg "attempt to decode some reasonable first few bytes as if it was $configured_encoding, and if this fails, fallback to assuming binary"?

Dreamsorcerer:
Ah, it looks like I made the changes through Github, not locally. I'll try cherry picking to a new branch.

Dreamsorcerer:
>     * however I'm wondering if there's anything cleaner/smarter we can do within the "number of bytes to read" helper instead (per your original 'fix' for this elsewhere)

That's a different issue, so not sure that function has anything to do with this one.

>     * or, since the "it's binary, please do not do any edge encoding/decoding" approach is arguably more correct, is there anything smarter we can do to autodetect?
>       
>       * are there approaches other tools use when handling unknown stdin?
>       * eg "attempt to decode some reasonable first few bytes as if it was $configured_encoding, and if this fails, fallback to assuming binary"?

To be honest, I never fully understood the reason decoding to a str and back again happens. So, you'll probably just have to make a decision on this one.

The only related thing I can think of is charset-normalizer, which can guess the encoding used. I'm still not sure I'd use it to try and detect if something is binary or not though. It's always going to be an estimate, so you'll likely still have edge cases that cause problems on occasion.

Dreamsorcerer:
Let me know if you want to continue this way, and I'll sort out the typing. It'll be a little tricky now, as the data being passed around is `str | bytes` and we're detecting it based on an option, rather than an isinstance() check.

invoke/runners.py
```diff
@@ -738,24 +778,32 @@ def write_our_output(self, stream: IO, string: str) -> None:
...
 
         :returns: ``None``.
 
         .. versionadded:: 1.0
         """
+        if isinstance(string, bytes) and hasattr(stream, "buffer"):
+            # Text streams such as sys.stdout cannot accept raw bytes; use
+            # the underlying binary buffer when there is one.
+            stream.buffer.write(string)
+            stream.buffer.flush()
```
Dreamsorcerer:
```suggestion
            stream.buffer.write(string)  # type: ignore[attr-defined]
            stream.buffer.flush()  # type: ignore[attr-defined]
```

invoke/runners.py
```diff
@@ -843,7 +899,7 @@ def read_our_stdin(self, input_: IO) -> Optional[str]:
                     raise
             # Decode if it appears to be binary-type. (From real terminal
             # streams, usually yes; from file-like objects, often no.)
-            if bytes_ and isinstance(bytes_, bytes):
+            if self.opts["decode_stdin"] and bytes_ and isinstance(bytes_, bytes):
```
Dreamsorcerer:
```suggestion
            if (
                self.opts["decode_stdin"]
                and bytes_
                and isinstance(bytes_, bytes)
            ):
```

tests/runners.py
```diff
@@ -415,6 +415,66 @@ def falls_back_to_defaultlocale_when_preferredencoding_is_None(self):
...
+            MockedStdin._write_proc_stdin = Mock()
+            self._run(
+                _,
+                klass=MockedStdin,
+                in_stream=BytesIO(payload),
+                decode_stdin=False,
+            )
+            written = b"".join(
+                args[0][0]
+                for args in MockedStdin.write_proc_stdin.call_args_list
```
Dreamsorcerer:
```suggestion
                for args in MockedStdin._write_proc_stdin.call_args_list
```

Dreamsorcerer:
@bitprophet Can we look at merging this now? I've updated it again and added some tests.

This is still a critical issue for me, and 3 years later, I'm still patching invoke to get something that is actually usable.

PR #916, opened 2023-02-12
Remaining typing tidyup

Dreamsorcerer:


Dreamsorcerer:
Think there's still some good changes here worth including, so just creating a PR with the remaining changes I had.

Will also push my partial typing of tests up later, if it looks like we can work around the awkward nested classes.

kuwv:
I recommend keeping 'if TYPE_CHECKING' as it reduces chances for circular imports and does not import unneeded modules at runtime.

Dreamsorcerer:
Yeah, I don't mind too much. I usually find it preferable to have real imports where possible.

Dreamsorcerer:
Kept one where the module was already imported for another object, have reverted the rest.

PR #922, opened 2023-02-27
Fix `Context.cd()` behavior to match the type hint

rouge8:
Issue found when fixing the type hint in typeshed: https://github.com/python/typeshed/pull/9823#pullrequestreview-1316129640

### DIFF
--- a/invoke/context.py
+++ b/invoke/context.py
@@ -389,7 +389,7 @@ class Context(DataProxy):
             (such as the various ``Path`` objects out there), and not just
             string literals.
         """
-        path = str(path)
+        path = os.fspath(path)
         self.command_cwds.append(path)
         try:
             yield


rouge8:
Ahh I see, reading the docstring it's the typehint that is wrong. Will fix.

rouge8:
Hmm, so the way I see to fix this would be to use a [`typing.Protocol`](https://docs.python.org/3/library/typing.html#typing.Protocol) with a `__str__` method, but it looks like invoke supports Python 3.6 which doesn't have that and doesn't currently have a dependency on `typing_extensions`. Should I add a dependency on `typing_extensions` to the dev requirements?

JelleZijlstra:
All objects have `__str__` so a protocol for `__str__` doesn't make sense. If you only want to support objects that are actually meant to be paths, `os.fspath` is a better choice.

kuwv:
I should have checked typeshed. I knew there was one for paramiko but I didn't know one was done for invoke. I'll see what I can use.

Thanks for the assist!

kuwv:
#956

Issue #860, opened 2022-05-13
Add pre & post tasks to help output


dflock:
It would be great if the help output for a task could include a list of pre & post tasks that this task calls. For example, if I have a `serve` task, that calls `clean` & `build` as pre-tasks, the help output could look something like this:

```console
❯ inv --help serve

Usage: inv[oke] [--core-opts] serve [other tasks here ...]

Docstring:
  Serve thing at http://localhost:$DOCS_PORT/ & live reload on changes. Default port: 5500

Options:
  none

Pre-tasks:
  Calls clean, build

Post-tasks:
  none
```

Invoke does know this information, and it would be great to make it visible to users so they know what running a task will do - without having to look at the `tasks.py` code - on manually having to add this text to the docstring.

neozenith:
Great suggestion. If you or anyone reading this has the bandwidth, opening a PR that references this issue and has appropriate test coverage would be much appreciated.

Our current priorities are working through dropping Python2 support to reduce our test matrix before addressing a backlog of issues.

PR #929, opened 2023-03-30
Add pre and post task lists into help output

tyzhnenko:
Example output

```bash
❯ inv --help main-task
Usage: inv[oke] [--core-opts] main-task [other tasks here ...]

Docstring:
  none

Options:
  none

Pre-tasks:
  pre-task
  pre-task-2

Post-tasks:
  none

```

Issue: #860

### DIFF
--- a/invoke/program.py
+++ b/invoke/program.py
@@ -781,6 +781,8 @@ class Program:
         ctx = self.parser.contexts[name]
         tuples = ctx.help_tuples()
         docstring = inspect.getdoc(self.collection[name])
+        pre_tasks = self.collection[name].pre
+        post_tasks = self.collection[name].post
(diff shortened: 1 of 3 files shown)


invoke/program.py
```diff
@@ -804,6 +806,27 @@ def print_task_help(self, name: str) -> None:
             print(self.leading_indent + "none")
             print("")
 
+        print("Pre-tasks:")
```
antoinebrl, line 809: Instead of displaying `none`, can the sections "Pre-tasks" and "Post-tasks" be made optional if the task doesn't define pre- and post- tasks respectively?

tyzhnenko: @antoinebrl sure, on the other hand, other sections have similar behavior with 'none' in the output. Won't it be confusing to have this difference?

Issue #937, opened 2023-05-02
Import is resolving the wrong module


fraenkel:
If you have a submodule under tasks that has the same name as a top-level module, the import will fail.

For example,  I have a module named common which is located in my site-packages.
My tasks directory looks like:
```
/project
  /tasks
    __init__.py
    common.py
     
```

The issue is pretty simple,
https://github.com/pyinvoke/invoke/blob/main/invoke/loader.py#L77

`import common` now comes from `root/tasks` instead of my site-packages.

fraenkel:
here is a simple adjustment to the current integration tests to show the failure.
```
diff --git a/integration/_support/package/tasks/invoke.py b/integration/_support/package/tasks/invoke.py
new file mode 100644
index 00000000..f9081c7d
--- /dev/null
+++ b/integration/_support/package/tasks/invoke.py
@@ -0,0 +1 @@
+HI = "hi"
diff --git a/integration/_support/package/tasks/module.py b/integration/_support/package/tasks/module.py
index 05f37ee5..ed2c259d 100644
--- a/integration/_support/package/tasks/module.py
+++ b/integration/_support/package/tasks/module.py
@@ -1,6 +1,7 @@
 from invoke import task
+from .invoke import HI


 @task
 def mytask(c):
-    print("hi!")
+    print(HI)
```

fraenkel:
Comparing 2.0 with 2.1.1, the main difference is sys.path.
2.0: `/invoke/integration/_support/package` is added
2.1.1: `/invoke/integration/_support/package/tasks` is added

PR #941, opened 2023-05-03
Add the correct path when it's a package

fraenkel:
When tasks is a package, adding itself to the sys.path causes a bit of trouble when you have collisions with top-level packages and modules within the package.


fraenkel:
Fixes #937

Issue #944, opened 2023-05-09
Invoke 2.1.0 / 2.1.1 is not able to read invoke.yaml configuration file


vaz-ar:
Hello,

I updated my virtualenv this morning and was faced with the following issue:

~~~bash
❯ inv git.checkout-main
Traceback (most recent call last):
  File "/home/arnaud/.virtualenvs/invoke/lib64/python3.11/site-packages/invoke/config.py", line 117, in __getattr__
    return self._get(key)
           ^^^^^^^^^^^^^^
  File "/home/arnaud/.virtualenvs/invoke/lib64/python3.11/site-packages/invoke/config.py", line 177, in _get
    value = self._config[key]
            ~~~~~~~~~~~~^^^^^
  File "/home/arnaud/.virtualenvs/invoke/lib64/python3.11/site-packages/invoke/config.py", line 168, in __getitem__
    return self._get(key)
           ^^^^^^^^^^^^^^
  File "/home/arnaud/.virtualenvs/invoke/lib64/python3.11/site-packages/invoke/config.py", line 177, in _get
    value = self._config[key]
            ~~~~~~~~~~~~^^^^^
KeyError: 'git'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/arnaud/.virtualenvs/invoke/bin/inv", line 8, in <module>
    sys.exit(program.run())
             ^^^^^^^^^^^^^
  File "/home/arnaud/.virtualenvs/invoke/lib64/python3.11/site-packages/invoke/program.py", line 398, in run
    self.execute()
  File "/home/arnaud/.virtualenvs/invoke/lib64/python3.11/site-packages/invoke/program.py", line 583, in execute
    executor.execute(*self.tasks)
  File "/home/arnaud/.virtualenvs/invoke/lib64/python3.11/site-packages/invoke/executor.py", line 140, in execute
    result = call.task(*args, **call.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/arnaud/.virtualenvs/invoke/lib64/python3.11/site-packages/invoke/tasks.py", line 138, in __call__
    result = self.body(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/arnaud/toolbox/work/python/invoke/tasks/git.py", line 53, in checkout_main
    for workspace in ctx.git.workspaces:
                     ^^^^^^^
[...]

Issue #948, opened 2023-05-19
types of context.config is not correct 


kasium:
The types of context.config is `dict[str, str]`, but should at least also support nested dicts in multiple levels. Also it can support any type as the value, so might just want to use `Any`

kuwv:
#956

kasium:
@kuwv @bitprophet any update?

kuwv:
Still waiting on Jeff. I need to make an update though. I'll take a look tonight.

PR #949, opened 2023-06-01
fix: resolve config regression with new loader capability

kuwv:
@bitprophet Seems there is an idiosyncrasy with how the new loader provides the parent path. I switched it to `cwd` but I'm sure there are some additional details here.

Issue is here: https://github.com/pyinvoke/invoke/issues/944

### DIFF
--- a/invoke/parser/context.py
+++ b/invoke/parser/context.py
@@ -91,7 +91,7 @@ class ParserContext:
         self.args = Lexicon()
         self.positional_args: List[Argument] = []
         self.flags = Lexicon()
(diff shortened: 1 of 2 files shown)


kuwv:
Resolve issue for typing within parser https://github.com/pyinvoke/invoke/issues/948.

jforcier-jumptrading:
Thanks for jumping on this, just hit it myself at work 😂  I'm not sure `cwd` is entirely accurate, if we want to mimic the old behavior, it needs to be "next to `tasks/`" (i.e. same as if it were a `tasks.py`). This feels/smells a lot like the previous bug related to the loader change, IIRC.

I'll have a poke shortly.

PR #956, opened 2023-07-10
test: update mypy typings

kuwv:


kuwv:
@bitprophet

kuwv:
@bitprophet do you think you can review this soon?

It should resolve multiple outstanding issues.

kuwv:
@Dreamsorcerer fysa

invoke/tasks.py
```diff
@@ -286,7 +287,7 @@ def get_arguments(
         return args
 
 
-def task(*args: Any, **kwargs: Any) -> Callable:
+def task(*args: Any, **kwargs: Any) -> "Task[T]":
```
Dreamsorcerer, line 290: The TypeVar is not being used correctly in this function. Either use `Task[Any]`, or if you want to spend a bit more time on it, you could try and type it properly.

I think the correct signature would look something along the lines of:
```
@overload
def task(arg: T, *, klass: Callable[[T]], U], **kwargs: Any) -> U:
    ...
@overload
def task(arg: T, **kwargs: Any) -> Task[T]:
    ...
@overload
def task(*args, klass: Callable[[Callable], T], **kwargs: Any) -> Callable[[Callable], T]:
    ...
@overload
def task(*args: Any, **kwargs: Any) -> Callable[[T], Task[T]]:  # T should actually be bound to Callable on this one.
    ...
```

But, frankly, it's rather complicated.

invoke/tasks.py
```diff
@@ -357,7 +358,7 @@ def inner(body: Callable) -> Task[T]:
         return _task
 
     # update_wrapper(inner, klass)
-    return inner
+    return cast('Task["T"]', inner)
```
Dreamsorcerer, line 361: Maybe this'll be fixed by the above comment, but if not, it'd be better to type ignore than cast here, given that it is literally not a Task being returned.

Dreamsorcerer (approved):

invoke/tasks.py
kuwv: This is following the recommendation provided by myotonic documentation: https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators

So, while it could be argued that it's not correct, it's still the recommended solution.

Dreamsorcerer: I think every example I see in that documentation matches my suggestions, not the original code.

The mistake is that a TypeVar is being used as a return type, but does not appear in the parameters (or anywhere else), which means that it is not doing the job of a TypeVar. The point of a TypeVar is to infer the precise type based on context. e.g. `def foo(a: _T) -> _T:` means that the return type will be the same as the input type. If you only put it in one place, then this gives no information at all (it wouldn't surprise me if mypy were to produce a type error in future, due to this being obviously incorrect usage).

Dreamsorcerer: Seems that mypy already produces an error for this, it's a bug that it didn't trigger for you:
https://github.com/python/mypy/issues/16113

invoke/tasks.py
kuwv: This is following the recommendation provided by mypy documentation: https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators

Dreamsorcerer: Right, I see what you're pointing to. It's not actually recommended, but maybe needed if we don't use typing-extensions to get ParamSpec.

Dreamsorcerer: No, wait, my point was that this is not a Task. The docs you point to are accepting a function and returning a function, the cast is to the function type. This is returning a function and then you're casting it as a Task.

docs:
```
F = TypeVar('F', bound=Callable[..., Any])
def printing_decorator(func: F) -> F:
    return cast(F, wrapper)
```

here:
```
def task(*args: Any, **kwargs: Any) -> Task[T]:
    def inner(body: Callable) -> Task[T]:
        ...
    return cast(Task[T], inner)  # inner is clearly not a Task
```

joaonc:
Would love to see this merged.   Per https://bitprophet.org/projects
> **Upcoming**
> 2024.07-8: will likely focus on Invoke features and major bugfixes

So maybe that will happen soon 🤞

Silverteal:
Not sure if this is the right place, but `invoke.runners.Runner.run()` may return Promise (when setting `asyncronous` to `True`), which is not in `run()`'s type hint, causing IDE to report false warnings.

kuwv:
@bitprophet do you have time to look at this patch?

bitprophet:
I'm incidentally taking a look at this and related things (my dayjob the last ~year has involved a lot of type hints, and Invoke's do seem to need a refresh).

Not sure if I'll get to everything this patch relates to or if I am on board with 100% of the changes, but I'll update it and the others once I've got something floated.

PR #957, opened 2023-07-11
Default run.shell to /bin/sh

tobim:
`/bin/bash` is not availabe on some popular distributions, for example NixOS. Defaulting to the more conservative `/bin/sh` improves portability.


kuwv:
This should be the correct default.

hydrocat:
Based on the CI results, the test cases are strictly requiring the shell to be `bash`. We could do like windows and use the `SHELL` environment variable, which would make sense for NixOS but would perhaps crash depending on the user's own configuration (for example, if they use fish, Oil, zsh, csh .. )

Maybe, we could query the system for bash's path.

tobim:
Well, you could use `/usr/bin/env` to find bash in `$PATH`.

kuwv:
BASH includes capabilities you don't need with Python/invoke. The fact that it's not included with Alpine or other Busybox-based distros is problematic.

The default should be '/bin/sh' because it's the right combination of capabilities and compatibility for invoke.

hydrocat:
Then, the tests should be updated

PR #981, opened 2024-01-02
Explicitly re-export names from top-level package using `__all__`

mthuurne:
This avoids the re-exports from being flagged as errors when a project that uses Invoke is checked using `mypy --no-implicit-reexport`.

It also obsoletes all flake8 suppressions in `__init__.py`.

I assumed the following names should be re-exported:
- every imported name that wasn't used in `__init__.py` itself
- `context.Context` and `runners.Result`

I assumed the following names should **not** be re-exported:
- `typing.Any` and `typing.Optional`

### DIFF
--- a/invoke/__init__.py
+++ b/invoke/__init__.py
@@ -1,10 +1,55 @@
+__all__ = (
+    "AmbiguousEnvVar",
+    "Argument",
+    "AuthFailure",
+    "Call",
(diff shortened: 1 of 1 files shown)


ogenstad:
Was also going to open a PR for this. I'd move the `__all__` so it's below the imports though.

PR #983, opened 2024-01-20
Speed up reading from in_stream

### LINKED ISSUE #774: Reading from in_stream is extremely slow by default
I've run into this using fabric2, but it seems to be an invoke issue: run passes data from in_stream to the process in an extremely slow manner. About 90 CPS (yes, that is characters per second). So if I enable `echo_stdin`, it looks like a screen from your typical hacker movie (only without the chirp).

Now while writing this bug report, I found a hint in the code (and then in the documentation) that this was intentional: `input_sleep` is set to 0.01 on invoke.Runner. But this seems like a pretty surprising default and, at least compared to the effect and the surprise factor, not a very well documented one. Also, probably my fault, but I did not find a way how to set it or disable it. After skimming the documentation, I have no idea which key to set. 

I've figured out how to set echo_stdin using the environment variables when running things from fabric, but the same scheme did not work out for input_sleep. Which means I wasn't able to guess the correct key. I've tried `runners.local.input_sleep` and `run.input_sleep. (So I ended up importing the class and then assigning 0 to the member but that's pretty lame.)

Actually, what's the point of this at all? If the input is not ready, the read will block anyway. If it's ready, why slow things down?

meshy:
By only sleeping when the input stream is waiting, this change ensures that we don't delay when there is still data available to read.

This provides a significant speed improvement to scripts which are passing a populated stream of data, rather than awaiting user input from stdin.

I had to modify an existing test to ensure that it continued to see `sleep`s.

I have also changed the default chunk read size to match [Python's default](https://docs.python.org/3/library/io.html#io.DEFAULT_BUFFER_SIZE), which is the "[“Preferred” blocksize for efficient file system I/O.](https://docs.python.org/3/library/os.html#os.stat_result.st_blksize)" (Of course, I'm happy to revert this if it makes the PR more palatable.)

Fixes https://github.com/pyinvoke/invoke/issues/774

### DIFF
--- a/invoke/runners.py
+++ b/invoke/runners.py
@@ -912,8 +912,10 @@ class Runner:
                 # race conditions re: unread stdin.)
                 if self.program_finished.is_set() and not data:
                     break
+                # When data is None, we're waiting for input on stdin.
                 # Take a nap so we're not chewing CPU.
-                time.sleep(self.input_sleep)
+                if data is None:
+                    time.sleep(self.input_sleep)
 
     def should_echo_stdin(self, input_: IO, output: IO) -> bool:
         """
--- a/invoke/runners.py
+++ b/invoke/runners.py
@@ -1,5 +1,6 @@
 import errno
 import locale
+import io
 import os
 import signal
 import struct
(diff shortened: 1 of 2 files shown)


meshy:
I've noticed some activity on this repo, and that there was a merge conflict on this branch, so I have rebased onto the latest state of the `main` branch and resolved the conflict.

PR #1022, opened 2025-02-09
Add command suggestions feature

krepysh:
# :rocket: Description

### :information_desk_person: Why is this change needed?

**This update enhances Invoke’s error handling by providing command suggestions when a command is mistyped, similar to how Git and Django’s `manage.py` handle such errors.**

This feature is especially useful for projects with complex command structures, particularly those involving multiple namespaces and deep nesting. In such cases, offering quick suggestions from the CLI tool significantly improves the user experience and efficiency.

### 🧑‍💻  Technical details

I implemented pattern matching using the `difflib` module, which is part of Python’s standard library, so no additional dependencies are required. 

Additionally, I introduced a new argument, `--suggestions` (or `-S`), which defaults to `False` to ensure backward compatibility. Ideally, this argument could be set to `True` by default in the future for an improved user experience.


invoke/program.py
```diff
@@ -86,6 +87,12 @@ def core_args(self) -> List["Argument"]:
                 default=False,
                 help="Echo executed commands before running.",
             ),
+            Argument(
+                names=("suggestions", "s"),
+                kind=bool,
+                default=False,
```
krepysh, line 93: Is it possible to default here to `True`?

If someone wanted to disable new feature, it is possible with `--no-suggestions` I guess.

invoke/program.py
```diff
@@ -985,3 +997,23 @@ def print_columns(
...
         print("")
+
+    def _possible_commands_msg(self, unknown_cmd: str):
+        try:
+            all_tasks = self.scoped_collection.task_names
+        except AttributeError:
+            all_tasks = {}
+
+        possible_cmds = list(all_tasks.keys())
+        suggestions = difflib.get_close_matches(
```
krepysh, line 1008: I am also working on a bit more advanced version of the string matching, more suits to `namespace.command` situation.

invoke/program.py
```diff
@@ -86,6 +87,12 @@ def core_args(self) -> List["Argument"]:
                 default=False,
                 help="Echo executed commands before running.",
             ),
+            Argument(
+                names=("suggestions", "s"),
+                kind=bool,
+                default=True,
```
krepysh, line 93: Is it possible to default here to True?

If someone wanted to disable new feature, it is possible with --no-suggestions I guess.

PR #1024, opened 2025-02-13
Add Python 3.12 and 3.13 to test matrix

jamesmyatt:
Note that Python 3.7 and 3.8 are EOL, so should probably be removed from the matrix


jamesmyatt:
I think the CI runners need updating because I think it requires pip >= 23.2.1 for Python 3.12: https://github.com/pypa/pip/issues/11501#issuecomment-1639513037

PR #1028, opened 2025-04-03
Remove support for EOL python versions

twmr:


twmr:
Does anyone know how I can switch the default python interpreter using in the CI from python3.6 to python 3.9?

PR #1034, opened 2025-05-28
Suggest invoke is more polite with errors.

macetw:
We ought be more polite and professional to typos.


macetw:
@jbbqqf could you support merging this pr?

Issue #1038, opened 2025-06-11
_pty_size does not handle SystemError (python 3.14)


kasium:
With python3.14 `result = fcntl.ioctl(sys.stdout, termios.TIOCGWINSZ, buf)` can also raise a SystemError which is not handelnd. Please add it to the except clause

https://github.com/pyinvoke/invoke/blob/2a8f95fa50e73f1520f8eb0463a8e51321290d98/invoke/terminals.py#L95-L112

PR #1074, opened 2026-06-18
Fix #1070: bytes_to_read buffer size for Python 3.14

### LINKED ISSUE #1070: `bytes_to_read()` raises `SystemError: buffer overflow` on Python 3.14 (`FIONREAD` undersized buffer)
Hi,

Same root cause as #1038 (fixed for `_pty_size`) but in `terminals.bytes_to_read()`, which was not touched by that fix.

On Python 3.14, `fcntl.ioctl` rejects undersized mutable buffers (cpython#144206). `bytes_to_read` passes a 2-byte buffer to `FIONREAD`, which writes a C int (4 bytes):

```
fionread = fcntl.ioctl(input_, termios.FIONREAD, b"  ")
return int(struct.unpack("h", fionread)[0])
```

which worked by accident on ≤3.13 thanks to the 1024-byte static copy buffer

*Repro*: any `Connection.run()` from a TTY-attached process on 3.14 crashes the `handle_stdin` thread

Do you want me to make a PR a fix with something like `fionread = fcntl.ioctl(input_, termios.FIONREAD, b"\x00\x00\x00\x00")` in https://github.com/pyinvoke/invoke/blob/main/invoke/terminals.py#L246-L247

Thx.

RachelXiaolan:
Fixes pyinvoke/invoke#1070.

## Bug

On Python 3.14, fcntl.ioctl rejects undersized mutable buffers (cpython#144206). `bytes_to_read()` was passing a 2-byte buffer to FIONREAD, which writes a C int (typically 4 bytes). Same root cause as the bug fixed in #1038 for `_pty_size`, but `bytes_to_read` was missed.

Worked by accident on <=3.13 thanks to CPython's 1024-byte static copy buffer. On 3.14, `Connection.run()` from a TTY-attached process crashes with SystemError: buffer overflow in `handle_stdin`.

## Fix

Pass a 4-byte buffer and unpack as a signed int:

```python
fionread = fcntl.ioctl(input_, termios.FIONREAD, b"\x00\x00\x00\x00")
return int(struct.unpack("i", fionread)[0])
```

## Tests

`tests/terminals/test_bytes_to_read.py` — 3 cases:
- buffer passed to ioctl is 4 bytes
- result is the int parsed from the 4-byte C int
- FIONREAD returning 0 is correctly handled
- non-TTY fallback still returns 1

Refs: AI-2138


hefee:
The test file look very complicated and not useful.

RachelXiaolan:
Thanks for the feedback @hefee! You're right — I overcomplicated it.

I've removed the standalone test file and added the test cases directly to the existing `tests/terminals.py`, matching the project's existing style (class-based, `@patch` decorators). The new tests are:

- `uses_4_byte_buffer_for_FIONREAD` — verifies ioctl gets a 4-byte buffer and the result is correct
- `returns_0_when_FIONREAD_is_zero` — edge case for zero bytes available

Both follow the same pattern as the existing `pty_size` and `bytes_to_read_` tests.

hefee:
Much better :) Unfortunately there are 3 pull requests try to fix the same issue, so nobody has looked if others already have started to fix and try to improve one pr:
#1053 , #1074, #1077
from the style of the MR I would say: #1074 and #1077 are created with the help of AI.

Issue #1, opened 2012-04-02
CLI task arguments


bitprophet:
See also: [Fabric #69](https://github.com/fabric/fabric/issues/69)
## Problem

Need to allow parameterization of tasks as invoked via the CLI (tasks invoked via Python are simply passing regular args/kwargs or their list/dict counterparts.) Sub-problem: translating strings/flag existence into Python types such as lists, booleans etc.

Current implementation in Fabric is `fab taskname:positional_arg_value,keyword_arg=kwarg_value,...`. Some other tools use "regular" CLI flags, e.g. `paver taskname -F value`.
### "Custom"/"attached" args

Pluses:
- **Works great for multi-task invokes**, zero ambiguity about which param values go to which tasks
- **Supports positional arguments** well

Minuses:
- **Requires custom parsing**, which may lead to bugs and at the very least increases code maintenance burden
- **Slight documentation burden** -- requires users to learn how that custom parsing works, re: need for escaping (esp. how that meshes with normal shell escapes) and etc.
### "Regular" flag args

Pluses:
- Can (probably) **just use argparse**
- Many **users already familiar** with it

Minuses:
- **Doesn't mesh well with multi-task invokes** unless we:
  - enforce BSD-style "order matters" (not even sure argparse can do this? EDIT: perhaps with subcommands, if >1 subcommand can be invoked at a time)
  - or use an ugly "task delimiter" syntax like `invoke task1 --arg1 --next task2 --arg2` (again, can argparse do this?)
  - or hope/pray that all simultaneously-invoked tasks have zero overlap in their kwargs and no positional args are ever needed.
    - How common is this? If we assume flags nix any possibility of positional args and that this is OK, it _could_ be reasonable to assume two or three tasks called together would all have different names for their kwargs. Maybe.
- **Namespace clashes with core** (non task related) flags such as `-c`/`--collection`
  - This also requires order-matters, or simply limiting the user-allowed namespace
  - Limiting the namespace also presents backwards compatibility problems -- if version 1.1 lacks a `--foo` flag and a user declares `mytask(foo='bar')`, and then we unknowingly add `--foo` to core in 1.2, that user's code will break.
- **May require more setup on the user side**, re: explicitly declaring each task's args/kwargs via decorators or whatnot
  - Though we could introspect the function signature and automatically create some, if needed
- **Doesn't allow for positional arguments** unless we go the delimiter route.
## Potential solutions
- Continue using Fabric's "custom parsing" solution pretty much as-is
[...]

PR #1078, opened 2026-07-25
Fix --help being swallowed as positional arg data after a task name

### LINKED ISSUE #982: `--help` after the command treats `--help` as positional argument
I have a repro where `--help` is incorrectly treated as a positional arg instead of triggering the invoke command help printout.
https://github.com/scottopell/invoke-help-bug

I've read through #205 and from what I understand this is not expected, `--help` and other core flags should be parsed out before task execution from what I understand.

Are there any suggested workarounds? Or is this unavoidable when the task takes positional arguments?

agu2347:
Fixes #982.

`invoke sometask --help`, where `sometask` has required positional arguments that haven't been supplied yet, incorrectly ran `sometask` with one of those arguments set to the literal string `"--help"`, instead of printing `sometask`'s help text. This was actually two separate bugs stacked together:

**1.** In `Parser.handle()`'s per-token dispatch, the "context still needs positional args, eat this token as one" branch was checked *before* the "this token is the core `--help` flag, given in a per-task position" branch. So whenever a task still had unfilled required positional args, `--help` got consumed as if it were one of those args' values, rather than being recognized as the help flag at all.

Moved the `--help`-specific check to run before the positional-arg branch. Unlike other core flags (still checked after positional args, preserving the deliberate existing behavior where a posarg's value may legitimately collide with another task's name), `--help` is never a value any task would legitimately want as positional input, so there's no equivalent ambiguity to preserve for this one flag specifically.

**2.** Even after fixing (1), `Parser.complete_context()` has its own, separate "did all positional args get filled?" validation that runs when finalizing a context, and it had no awareness of whether `--help` had been requested for that context. So even once `--help` was no longer being swallowed as posarg data, this second check would still immediately raise `"'sometask' did not receive required positional arguments: ..."`, before `Program.parse_cleanup()`'s already-correct `--help`-printing logic (which checks the exact same flag value set in fix (1)) ever got a chance to run.

Made `complete_context()` skip its missing-positional-args check when `--help` has already been set to this context's name, using the same signal `Program.parse_cleanup()` already relies on downstream.

**Testing**: verified via `Program.run()` against a real task collection with a required-positional-arg task: confirmed `punch --help` (where `punch` requires two positional args) printed `"Hello, --help!"`-style incorrect output before either fix, then raised a spurious missing-args `ParseError` after only fix (1), and only correctly prints `punch`'s help text with both fixes applied. Also verified: bare `--help` (no task) still works; `--help` before the task name still works (pre-existing, different code path); tasks with no positional args and `--help` are unaffected; tasks with multiple positional args and `--help` work correctly; and genuinely missing positional args (without `--help`) still correctly raise the intended error.

Added a regression test alongside the existing (differently-ordered) per-task `--help` test, using the same `punch` fixture task and `expect()` test helper. I confirmed it fails with the original code and passes with both fixes applied. Ran the full existing parser/program/CLI test suites (286 passed: 285 baseline + 1 new, 3 skipped) and the entire project test suite with an exact before/after comparison (114 failed / 859 passed / 11 skipped identically with and without this change -- confirmed those failures are pre-existing sandbox limitations around subprocess/PTY execution in `tests/runners.py`, unrelated to this change).


Mukller (approved):
Verified locally on the PR branch vs upstream master (Python 3.13, Windows) with a minimal task file:

```python
@task
def build(ctx, target): print(f"BUILDING {target}")
```

```
$ invoke build --help

master  (before): prints "BUILDING --help"   <- --help silently consumed as
                                                the required positional value;
                                                task actually executes
branch  (after): prints build's usage/help, exit 0
```

This is the exact GH #982 scenario, and the reproduction shows it's worse than a missing-help annoyance: on master the task *runs* with `--help` as argument data.

The two-part fix reads correctly:

1. `handle()` checks the initial-context `--help` flag *before* the positional-arg branch — justified by the comment's argument that no task legitimately wants `--help` as positional data (no ambiguity to preserve), unlike other core flags.
2. `complete_context()` skips the "did not receive required positional arguments" error when help was requested for this context — otherwise viewing help for a posarg-taking task would be impossible even after fix #1.

Changelog entry references :bug:`982` properly; new tests added in tests/program.py (POSIX-gated via `termios`, so I couldn't execute them locally on Windows — but the functional repro above exercises the same code path end-to-end). Everything else green on CI.

PR #1080, opened 2026-07-30
fix: raise ConfigFileNotFound when explicit runtime config path is missing (#560)

### LINKED ISSUE #560: Invalid runtime config file paths don't raise any error
What it says on the tin. I was writing some tests around #556 and typoed the runtime config file name; got no error about it whatsoever. Feels bad.

Mukller:
## Summary

Fixes #560 — when the user provides an explicit runtime config path (via `--config /path/to/file.yaml` or `runtime_path=` kwarg) and that file does not exist, invoke silently continues with no config loaded and no error.

## Root cause

`Config._load_file()` catches `IOError` with `errno == 2` (file not found) and logs a debug message, then continues. This is correct for the *implicit* config search (invoke tries `.invoke.yaml`, `.invoke.yml`, etc. in order and silently skips each missing option). But for the **explicit runtime config path** (`absolute=True`), the user chose a specific file; a missing file is a user error, not "not configured".

## Fix

Three changes:

**`invoke/exceptions.py`** — new `ConfigFileNotFound(IOError)` exception class with a `.path` attribute.

**`invoke/config.py`** — in the `errno == 2` handler, `raise ConfigFileNotFound` when `absolute=True`. Non-runtime sources keep their silent-skip behaviour.

**`tests/config.py`** — two regression tests:
- `missing_runtime_path_raises_config_file_not_found`
- `missing_runtime_path_error_includes_path`

## Before / after

```
# Before: runs normally, typo in path goes undetected
$ inv --config /tmp/doesnt_exist.yaml sometask

# After:
$ inv --config /tmp/doesnt_exist.yaml sometask
invoke.exceptions.ConfigFileNotFound: Runtime config file '/tmp/doesnt_exist.yaml' was not found. Check the path for typos.
```

`ConfigFileNotFound` extends `IOError` so existing `except IOError` handlers in user code continue to work.

## Checklist

- [x] New exception class in `exceptions.py` with `.path` attribute
- [x] `config.py` raises `ConfigFileNotFound` only when `absolute=True`
- [x] Implicit config search (system/user/project) unaffected
- [x] Regression tests added
- [x] `ConfigFileNotFound` importable from `invoke.exceptions`

### DIFF
--- a/invoke/__init__.py
+++ b/invoke/__init__.py
@@ -8,6 +8,7 @@ from .exceptions import (  # noqa
     AmbiguousEnvVar,
     AuthFailure,
     CollectionNotFound,
+    ConfigFileNotFound,
     CommandTimedOut,
     Exit,
     ParseError,
--- a/invoke/config.py
+++ b/invoke/config.py
@@ -9,7 +9,9 @@ from types import ModuleType
 from typing import Any, Dict, Iterator, Optional, Tuple, Type, Union
 
 from .env import Environment
-from .exceptions import ConfigFileNotFound, UnknownFileType, UnpicklableConfigMember
+from .exceptions import (
+    ConfigFileNotFound, UnknownFileType, UnpicklableConfigMember
+)
 from .runners import Local
 from .terminals import WINDOWS
 from .util import debug, yaml
@@ -900,7 +902,8 @@ class Config(DataProxy):
             except IOError as e:
                 if e.errno == 2:
                     # Absolute paths (runtime config) were explicitly given by
-                    # the user; a missing file is an error, not "not configured".
+                    # the user; a missing file is an error,
+                    # not "not configured".
                     if absolute:
                         raise ConfigFileNotFound(str(filepath)) from e
                     err = "Didn't see any {}, skipping."
--- a/invoke/config.py
+++ b/invoke/config.py
@@ -9,7 +9,7 @@ from types import ModuleType
 from typing import Any, Dict, Iterator, Optional, Tuple, Type, Union
 
 from .env import Environment
-from .exceptions import UnknownFileType, UnpicklableConfigMember
+from .exceptions import ConfigFileNotFound, UnknownFileType, UnpicklableConfigMember
 from .runners import Local
 from .terminals import WINDOWS
 from .util import debug, yaml
--- a/invoke/config.py
+++ b/invoke/config.py
@@ -899,6 +899,10 @@ class Config(DataProxy):
             # Typically means 'no such file', so just note & skip past.
             except IOError as e:
                 if e.errno == 2:
+                    # Absolute paths (runtime config) were explicitly given by
+                    # the user; a missing file is an error, not "not configured".
+                    if absolute:
+                        raise ConfigFileNotFound(str(filepath)) from e
                     err = "Didn't see any {}, skipping."
                     debug(err.format(filepath))
                 else:
--- a/invoke/config.py
+++ b/invoke/config.py
@@ -10,7 +10,9 @@ from typing import Any, Dict, Iterator, Optional, Tuple, Type, Union
 
 from .env import Environment
 from .exceptions import (
-    ConfigFileNotFound, UnknownFileType, UnpicklableConfigMember
+    ConfigFileNotFound,
+    UnknownFileType,
+    UnpicklableConfigMember,
 )
 from .runners import Local
 from .terminals import WINDOWS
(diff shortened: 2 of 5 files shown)


Mukller (commented):
## Code review

### Why `absolute=True` is the right guard

`Config._load_file` is called for five config tiers: system, user, project, runtime, and `shell_env`. Only the *runtime* tier passes `absolute=True` (line 893 in `Config.load_runtime()`). All other tiers use relative paths found by the implicit search, where a missing file is expected behaviour.

Adding the `if absolute:` check is therefore surgical: it fires precisely when the user has explicitly provided a path (via `--config` or `runtime_path=`), and leaves the implicit search completely untouched.

### `ConfigFileNotFound` extends `IOError`

This matters for backward compatibility: existing user code that wraps invoke calls with `except IOError` will still catch the new exception without any changes. The `.path` attribute provides the path that was requested, useful for producing a clear error message downstream.

### Test coverage

The two new tests in `tests/config.py` mirror the style of the existing `unknown_suffix_in_runtime_path_raises_useful_error` test (uses `@raises` from `pytest_relaxed`, creates a `Config` with `runtime_path=`, calls `load_runtime()`). The second test uses `pytest.raises` as a context manager to inspect `exc_info.value.path`, confirming the attribute is set correctly.

### No change to the debug path

When `absolute=False` (all implicit tiers), the `errno == 2` handler still logs at debug level and continues — exactly as before. There is no regression risk for users who don't supply an explicit runtime path.

### Suggested follow-up (non-blocking)

The CLI (`program.py`) catches `ConfigFileNotFound` as an `IOError` at the top level and prints the traceback. A small enhancement would be to catch `ConfigFileNotFound` explicitly in the CLI's error handler to print only the message (no traceback), matching how `UnknownFileType` is handled in `program.py`. That is out of scope for this PR.

Mukller (commented):
## Code review

### Why `absolute=True` is the right guard

`Config._load_file` is called for five config tiers: system, user, project, runtime, and `shell_env`. Only the *runtime* tier passes `absolute=True` (line 893 in `Config.load_runtime()`). All other tiers use relative paths found by the implicit search, where a missing file is expected behaviour.

Adding the `if absolute:` check is therefore surgical: it fires precisely when the user has explicitly provided a path (via `--config` or `runtime_path=`), and leaves the implicit search completely untouched.

### `ConfigFileNotFound` extends `IOError`

This matters for backward compatibility: existing user code that wraps invoke calls with `except IOError` will still catch the new exception without any changes. The `.path` attribute provides the path that was requested, useful for producing a clear error message downstream.

### Test coverage

The two new tests in `tests/config.py` mirror the style of the existing `unknown_suffix_in_runtime_path_raises_useful_error` test (uses `@raises` from `pytest_relaxed`, creates a `Config` with `runtime_path=`, calls `load_runtime()`). The second test uses `pytest.raises` as a context manager to inspect `exc_info.value.path`, confirming the attribute is set correctly.

### No change to the debug path

When `absolute=False` (all implicit tiers), the `errno == 2` handler still logs at debug level and continues — exactly as before. There is no regression risk for users who don't supply an explicit runtime path.

### Suggested follow-up (non-blocking)

The CLI (`program.py`) catches `ConfigFileNotFound` as an `IOError` at the top level and prints the traceback. A small enhancement would be to catch `ConfigFileNotFound` explicitly in the CLI's error handler to print only the message (no traceback), matching how `UnknownFileType` is handled in `program.py`. That is out of scope for this PR.

Mukller:
Fixed multiple CI failures in this push:

**Lint (flake8 E501)**: Four lines exceeded the 79-character limit. Wrapped the long import, comments, and string to satisfy both flake8 and Black.

**Style check (Black)**: Applied Black-compatible formatting throughout.

**Test failures** (3 tests):
- `nonexistent_files_are_skipped_and_logged`: This test captured the pre-fix behavior (silent skip for missing runtime path). Updated it to `@raises(ConfigFileNotFound)` to reflect the new intended behavior.
- `Init::exposes_bindings::exceptions`: `ConfigFileNotFound` was not exported from `invoke`'s top-level `__init__.py`. Fixed by adding it to the exports.
- `tasks_dedupe_honors_configuration`: The test used `-f dedupe.yaml` but `dedupe.yaml` did not exist in the test fixtures, relying on the silent-skip behavior this PR intentionally removes. Created `tests/_support/configs/dedupe.yaml` with `tasks: dedupe: true` (the opposite of `no-dedupe.yaml`) so the "flag beats runtime" assertion is meaningful.

PR #1087, opened 2026-09-18
Fix catastrophic performance

### LINKED ISSUE #819: Awful performance due to reading bytes 1 at a time
I'm using fabric to send a stream of binary data (can be GBs in size). The invoke code reads the data 1 byte at a time. This seems to result in an upload speed maxing out at 10KB/s, which would take days to upload the necessary data.

Perhaps there can be a way to customise the chunk size for reading?

I easily achieved over 2MB/s after changing that number (atleast 200x improvement, probably more if I had a faster internet connection):
```diff
--- a/invoke/runners.py
+++ b/invoke/runners.py
@@ -790,13 +790,13 @@ class Runner(object):
         # read instead of once per session, which could be costly (?).
         bytes_ = None
         if ready_for_reading(input_):
-            bytes_ = input_.read(bytes_to_read(input_))
+            bytes_ = input_.read(2**20)
             # Decode if it appears to be binary-type. (From real terminal
             # streams, usually yes; from file-like objects, often no.)
```

Dreamsorcerer:
Fixes #819.

Without patching invoke, no matter how much hardware is thrown at it, it will never exceed a throughput of 100 B/s (yes, bytes). This is a multi-pronged approach to fix the problem.

- When the output is not an actual terminal, use the chunk size instead of falling back to 1 byte chunks.
- If we received data, don't sleep for 10ms between chunks when there's probably more data to be processed.
- Increase the default chunk size to 64 KiB, which better matches other typical default sizes.
- Add a chunk size parameter to allow the user to use larger chunk sizes.

The first 3 points improve performance from 100 B/s to atleast comfortably over 10 MiB/s. Using a custom 2 MiB chunk size via point 4, I was able to get an extra ~6% performance on top of that.

-----

Combined with #915, this will finally allow me to stop patching invoke and just use it as installed library.


invoke/runners.py
```diff
@@ -830,8 +841,18 @@ def read_our_stdin(self, input_: IO) -> Optional[str]:
...
         bytes_ = None
         if ready_for_reading(input_):
+            if isatty(input_):
+                num_bytes = bytes_to_read(input_)
+                read = input_.read
+            else:
+                num_bytes = self.read_chunk_size
+                # read1 avoids blocking, but is not available on text streams.
+                read = (
+                    input_.read1 if hasattr(input_, "read1")
```
Dreamsorcerer:
```suggestion
                    input_.read1 if hasattr(input_, "read1")  # type: ignore[attr-defined]
```

invoke/runners.py
```diff
@@ -830,8 +841,18 @@ def read_our_stdin(self, input_: IO) -> Optional[str]:
...
+                    else input_.read
```
Dreamsorcerer:
```suggestion
                    input_.read1 if hasattr(input_, "read1") else input_.read  # type: ignore[attr-defined]
```

tests/runners.py
```diff
@@ -414,6 +414,95 @@ def falls_back_to_defaultlocale_when_preferredencoding_is_None(self):
                 fake_locale.getdefaultlocale.return_value = (None, None)
                 fake_locale.getpreferredencoding.return_value = "FALLBACK"
                 assert self._runner().default_encoding() == "FALLBACK"
+
```
Dreamsorcerer:
```suggestion

```

tests/runners.py
```diff
@@ -414,6 +414,94 @@ def falls_back_to_defaultlocale_when_preferredencoding_is_None(self):
                 fake_locale.getdefaultlocale.return_value = (None, None)
                 fake_locale.getpreferredencoding.return_value = "FALLBACK"
                 assert self._runner().default_encoding() == "FALLBACK"
+    class read_chunk_size:
```
Dreamsorcerer, line 418:
```suggestion

    class read_chunk_size:
```

invoke/runners.py
```diff
@@ -830,8 +841,17 @@ def read_our_stdin(self, input_: IO) -> Optional[str]:
...
         bytes_ = None
         if ready_for_reading(input_):
+            if isatty(input_):
+                num_bytes = bytes_to_read(input_)
+                read = input_.read
+            else:
+                num_bytes = self.read_chunk_size
+                # read1 avoids blocking, but is not available on text streams.
+                read = (
+                    input_.read1 if hasattr(input_, "read1") else input_.read  # type: ignore[attr-defined]
```
Dreamsorcerer:
```suggestion
                    input_.read1 if hasattr(input_, "read1") else input_.read  # type: ignore[attr-defined]  # noqa: E501
```

